PR #48, opened 2025-07-01
feat/async-mirror-node-key-type-detection

michalrozekariane:
## Description

This PR refactors key type detection logic throughout the codebase.
The old utility `detectKeyTypeFromString` was removed and replaced with a new asynchronous utility `detectKeyTypeFromMirrorNode`, which detects key type using the Hedera mirror node API.

## Related Issue

https://github.com/hedera-dev/hedera-agent-kit/issues/40

## Checklist

- [x] My code follows the contributing guidelines of this project.
- [x] I have performed a self-review of my code.
- [x] My commits are DCO-signed (`git commit -s`).
- [x] I have commented my code where needed, particularly in hard-to-understand areas.
- [x] I have added tests to prove my fix is effective or that my feature works.


README.md
```diff
@@ -125,7 +125,7 @@ async function main() {
       'HEDERA_ACCOUNT_ID and HEDERA_PRIVATE_KEY must be set in .env'
     );
 
-  const agentSigner = new ServerSigner(operatorId, operatorKey, network);
+  const agentSigner = await ServerSigner.create(operatorId, operatorKey, network);
```
kantorcodes, line 128: This will break in multiple places. 

1. The builder methods were designed to handle keys that are not the user's own keys, e.g. passed in input. I don't believe there is a mirror node call for this. The SDK handles automatic detection if it should use the user's own PK or a custom ones, and the builders can be used independently with direct input too. 

2. Passing in a private key arg would work for the signer itself, but not for input @jaycoolh for the above reason 

3. As @jaycoolh mentioned, it's a lot of async calls which adds more room for error.

Overall, this would reduce DX and I still advocate to fixing the original method to support all key type edge cases with integration tests rather.

kantorcodes (changes requested):

kantorcodes:
If there is a degree of urgency here, a more immediate fix would be to document that hex encoded ED25119 keys (plus other edge cases) are not supported yet. 

At any rate, my team is currently looking at this specific case and we should have an update soon.

README.md
kantorcodes: > Thought:
> 
> 1. we update the ServerSigner to use a `PrivateKey` and then remove dependency on checking for key type
> 2. for the account builder, we split create account functionality into two:
>    a) create ecdsa account (default to this if not specified and use `setECDSAKeyWithAlias()`*)
>    b) create ed25519 account
> 
> @kantorcodes for the create account tool to work, must the user use a key / alias in the prompt? Or is there a flow where a privkey is generated for them
> 
> *this is preferred / default way because it allows Hedera accounts to use all the EVM features e.g. ECRecover
> 
> cc @se7enarianelabs

1. When someone still has to do that kind of work, I tend to defer to "magic under the hood," to ease the developer experience, but that's my perspective. Your path forward is plausible. 
2. I agree the account builder should support both. There's lots of other cases in the builders where we need to keep the options open for custom keys though, like other entity creation. The `conversational-agent.ts` supports generating a PK on their behalf using their own Public Key (this is the default), when the prompt specifically provides a key, then it'll override that default.

Issue #68, opened 2025-07-16
Create min set of mirrornode query tools


jaycoolh:
get hbar balance
get token balance
read topic

PR #75, opened 2025-07-16
feat: Add support for non-fungible token (NFT) creation on Hedera

skurzyp-arianelabs:
## Description
This is draft PR with changed `ts-config` and project `type` set to `module` - this actions have been taken to resolve the import issues. The `ts-config` is a mix of the v3( I left the shortened imports) and v2 (`target`, `module` required alignment and I have copied rest of the `compilerOptions`)

This PR also introduces very simple creation of NFTs with minimal params.

## Related Issue

If this PR fixes a bug or resolves a feature request, link it here:
- Related to  #68 

## Checklist

- [X] My code follows the contributing guidelines of this project.
- [ ] I have performed a self-review of my code.
- [X] My commits are DCO-signed (`git commit -s`).
- [ ] I have commented my code where needed, particularly in hard-to-understand areas.
- [ ] I have added tests to prove my fix is effective or that my feature works.

### DIFF
--- a/src/examples/langchain.ts
+++ b/src/examples/langchain.ts
@@ -4,16 +4,17 @@ import type { ChatPromptTemplate } from "@langchain/core/prompts";
 import { pull } from "langchain/hub";
 import { AgentExecutor, createStructuredChatAgent } from "langchain/agents";
 import { Client, PrivateKey } from "@hashgraph/sdk";
-import { AgentMode } from "../shared/configuration";
+import { AgentMode } from "@/shared/configuration";
+import dotenv from "dotenv";
 
-require("dotenv").config();
+dotenv.config();
 
 const llm = new ChatOpenAI({
   model: "gpt-4o-mini",
 });
 
 const client = Client.forTestnet()
-//.setOperator(process.env.ACCOUNT_ID!, PrivateKey.fromStringED25519(process.env.PRIVATE_KEY!))
+  .setOperator(process.env.ACCOUNT_ID!, PrivateKey.fromStringED25519(process.env.PRIVATE_KEY!))
 
 const hederaAgentToolkit = new HederaAgentLangchainToolkit({
   client,
@@ -22,10 +23,13 @@ const hederaAgentToolkit = new HederaAgentLangchainToolkit({
       fungibleToken: {
         create: true,
       },
+      nonFungibleToken: {
+        create: true,
+      },
     },
     context: {
-      mode: AgentMode.RETURN_BYTES,
-      accountId: "0.0.123123",
+      mode: AgentMode.AUTONOMOUS,
+      // accountId: "0.0.123123", // TODO: accountId should be passed only if the AgentMode is non-custodial?
     },
   },
 });
@@ -49,9 +53,16 @@ const hederaAgentToolkit = new HederaAgentLangchainToolkit({
     returnIntermediateSteps: true
   });
 
+  // EXAMPLE PROMPT FOR FT CREATION
+  // const response = await agentExecutor.invoke({
+  //   input: `
+  //     Create a token called Hello World with symbol HELLO with treasury account 0.0.123123.
+  //   `,
+  // });
+
   const response = await agentExecutor.invoke({
     input: `
-      Create a token called Hello World with symbol HELLO with treasury account 0.0.123123. 
+      Create a nft token called Hello World with symbol HELLO.
     `,
   });
 
--- a/src/shared/hedera-utils/hedera-builder.ts
+++ b/src/shared/hedera-utils/hedera-builder.ts
@@ -1,6 +1,5 @@
-import { Client, TokenCreateTransaction } from "@hashgraph/sdk";
-import { Context } from "../configuration";
-import { createFungibleTokenParameters } from "../parameter-schemas/hts.zod";
+import { TokenCreateTransaction, TokenSupplyType } from "@hashgraph/sdk";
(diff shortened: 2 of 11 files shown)


src/examples/langchain.ts
```diff
@@ -3,23 +3,39 @@ import { ChatOpenAI } from '@langchain/openai';
...
+);
 
 const hederaAgentToolkit = new HederaAgentLangchainToolkit({
   client,
   configuration: {
     actions: {
       fungibleToken: {
         create: true,
+        airdrop: true,
+        transfer: true,
```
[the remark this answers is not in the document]
jaycoolh: My bad. This was a mistake from me copying and pasting over, didn't notice... we only need one

src/shared/hedera-utils/hedera-builder.ts
```diff
@@ -1,9 +1,110 @@
...
+  ) {
+    const tx = new TokenCreateTransaction()
+      .setTokenName(params.tokenName)
+      .setTreasuryAccountId(params.treasuryAccountId!)
+      .setTokenType(TokenType.FungibleCommon)
+      .setDecimals(params.decimals)
+      .setInitialSupply(params.initialSupply ?? 0)
+      .setSupplyType(
+        params.supplyType === 'infinite' ? TokenSupplyType.Infinite : TokenSupplyType.Finite,
+      );
```
jaycoolh, line 36: How come we don't use the params object and pass it into the constructor? I think this is a nice approach as then to extend functionality we only need to update the prompt, zod schema, and normaliser. We don't need to touch this builder

src/shared/hedera-utils/hedera-builder.ts
```diff
@@ -1,9 +1,110 @@
...
+      .setMaxSupply(maxSupply)
+      .setSupplyKey(PublicKey.fromString(params.supplyKey!));
+  }
+
+  static transferHbar(params: z.infer<ReturnType<typeof transferHbarParameters>>) {
+    return new TransferTransaction()
+      .addHbarTransfer(params.destinationAccountId, new Hbar(params.hbarAmount))
+      .addHbarTransfer(params.sourceAccountId as string, new Hbar(params.hbarAmount).negated())
+      .setTransactionMemo(params.transactionMemo || '');
+  }
```
jaycoolh, line 73: Same here - I think using the constructor args can help us future proof data structures when we look to extend in the future

```
{
  hbarTransfers: [
    {accountId: 'xxx', amount: n},
    {accountId: 'yyy', amount: -n},
  ]
}

Comment (jaycoolh) [reply src/shared/hedera-utils/hedera-builder.ts:73]: Would AI be able to generate this schema? Or at least, for the time being, our normaliser

File: src/shared/hedera-utils/hedera-builder.ts
Comment (jaycoolh) [inline src/shared/hedera-utils/hedera-builder.ts:1]: Overall on this one I think we should standardise our builder to use the interface required in the Hedera Transaction constructor.

We can keep our prompts etc the same. Then for the future we only need to update the prompt stuff and the normaliser to add new features to a specific transaction

@se7enarianelabs thoughts?

File: src/shared/tools/account/transfer-hbar.ts [hunk 4]
```diff
@@ -0,0 +1,52 @@
...
+import HederaParameterNormaliser from '@/shared/hedera-utils/hedera-parameter-normaliser';
+
+const transferHbarPrompt = (_context: Context = {}) => `
+This tool will transfer HBAR to an account.
+
+It takes three arguments:
+- hbarAmount (number): amount of hbar to transfer.
+- destinationAccountId (str): account to transfer hbar to.
+- transactionMemo (str, optional): optional memo for the transaction.
+`;
```
jaycoolh, line 17: Should our prompt have a message which takes the context if context.accountId exists to show that "context.accountId is referred to as my account"?

src/shared/tools/fungible-token/airdrop-fungible-token.ts
```diff
@@ -0,0 +1,61 @@
...
+This tool will airdrop a fungible token on Hedera.
+
+It takes five arguments:
+- tokenId (str): The id of the token.
+- amount (int): The amount of tokens to airdrop - given by user in display units.
+- sourceAccountId (str, optional): The account to airdrop the token from. If the user does not explicitly provide a value for this field, do not include it at all in the request — leave it undefined. Do not set it to null or any placeholder string.
+-  recipients (array): A list of recipient objects, each containing:
+  - accountId (string): The recipient's account ID (e.g., "0.0.1234").
+  - amount (number or string): The amount of tokens to send to that recipient (in base units).
+- transactionMemo (str, optional): optional memo for the transaction.
```
jaycoolh, line 21: I like this approach; I think we can do the same for hbar transfer

jaycoolh (approved):

src/shared/tools/account/transfer-hbar.ts
skurzyp-arianelabs: Probably should. Then I'll add the sourceAccountId here as the one of the tool parameters and if the accountId is in context it will be set and if not it will be left undefined and defined in the `hedera-parameter-normaliser` the following way:

```ts
    const sourceAccountId =
      params.sourceAccountId ?? context.accountId ?? client.operatorAccountId?.toString();
```

src/shared/hedera-utils/hedera-builder.ts
skurzyp-arianelabs: Ill take a closer look at it

PR #76, opened 2025-07-16
Feat/min set of query tools

### LINKED ISSUE #68: Create min set of mirrornode query tools

se7enarianelabs:
## Description

Describe the changes introduced by this PR. Include the context of the change, the reason behind it, and any relevant documents or screenshots.

## Related Issue

If this PR fixes a bug or resolves a feature request, link it here:
- Closes #68 

## Checklist

- [ ] My code follows the contributing guidelines of this project.
- [ ] I have performed a self-review of my code.
- [ ] My commits are DCO-signed (`git commit -s`).
- [ ] I have commented my code where needed, particularly in hard-to-understand areas.
- [ ] I have added tests to prove my fix is effective or that my feature works.


package.json
```diff
@@ -18,11 +18,16 @@
   "dependencies": {
     "@hashgraph/sdk": "^2.68.0",
     "@langchain/core": "^0.3.62",
+    "bignumber.js": "^9.3.1",
     "dotenv": "^17.2.0",
+    "web-streams-polyfill": "^4.1.0",
```
skurzyp-arianelabs, line 23: Where is this library used?

src/shared/tools/queries/get-account-token-balances-query.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export const getAccountTokenBalancesQueryPrompt = (_context: Context = {}) => `
+This tool will return the token balances for a given Hedera account.
+
+It takes two arguments:
+- accountId (str): The account ID to query.
+- tokenId (str, optional): The token ID to query for. If not provided, all token balances will be returned.
+`;
+
+export const getAccountTokenBalancesQuery = async (
```
jaycoolh, line 16: How do we handle the case of "get my token balance for 0.0.XXX

jaycoolh: There might be something in the v2 which we can learn from

src/shared/tools/queries/get-hbar-balance-query.ts
```diff
@@ -0,0 +1,48 @@
...
+    const mirrornodeService = getMirrornodeService(context.mirrornodeConfig!);
+    const balance: BigNumber = await mirrornodeService.getAccountHBarBalance(params.accountId);
+    return { accountId: params.accountId, hbarBalance: balance.toString() };
+  } catch (error) {
+    console.error('Error getting HBAR balance', error);
+    if (error instanceof Error) {
+      return error.message;
+    }
+    return 'Failed to get HBAR balance';
+  }
```
jaycoolh, line 32: shall we create a utility to handle this as I think we will be repeating it a lot?

jaycoolh (approved):

src/shared/tools/queries/get-account-token-balances-query.ts
skurzyp-arianelabs: I'm just working on similar case with setting the keys for FT/NFT/Topics and I think that the best approach would be injecting the data from `context` in to the prompt and then further validating the tool params created by the LLM in the `hedera-parameter-normaliser`.

```TS
const createFungibleTokenPrompt = (context: Context = {}) => `
context:
 - operator Account: ${context.accountId}. The user will refer to this as the "my account"
 - operator Account Public Key: ${context.accountPublicKey}. The public key of the operator account. Should be used as key if no other key is provided.
This tool will create a fungible token on Hedera.

It takes four arguments:
- tokenName (str): The name of the token.
- tokenSymbol (str, optional): The symbol of the token.
- initialSupply (int, optional): The initial supply of the token.
- supplyType (str, optional): The supply type of the token. Can be "finite" or "infinite". Defaults to "finite".
- maxSupply (int, optional): The max supply of the FT token. Can't be set if supplyType is "infinite". Defaults to 1000000 if set to "finite" and not passed.
- decimals (int, optional): The number of decimals of the token. Defaults to 0.
- treasuryAccountId (str, optional): The treasury account of the token. Defaults to the operator account.
- adminKey (str, optional): The admin key of the token.
`;
```

src/shared/configuration.ts
```diff
@@ -10,6 +12,12 @@ export type Actions = {
   balance?: {
     read?: boolean;
   };
+  accountQuery?: {
+    getAccountBalanceQuery?: boolean;
+    getAccountQuery?: boolean;
+    getAccountTokenBalancesQuery?: boolean;
+    getTopicMessagesQuery?: boolean;
+  };
```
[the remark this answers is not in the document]
se7enarianelabs: Will think about it

src/shared/tools/queries/get-account-token-balances-query.ts
jaycoolh: Cool - is a similar approach to how Michael did it in v2? sounds similar to @skurzyp-arianelabs point of view too

skurzyp-arianelabs: To be honest they flow of params extraction and parsing in v2 is still unclear for me. I can see that they were injecting some info to the zod schema descriptions e.g. ` ... Defaults to user's account if in user-centric context and not specified.`. That means the user account data is defined somewhere in the context of the agent/session. Also, some parsing is handled by the builders. The approach I have proposed is far simpler - we are just injecting data into the template.

Issue #66, opened 2025-07-16
Create example for return bytes flow with Langchain agent


jaycoolh:

b-l-u-e:
@jaycoolh please assign me this issue

PR #90, opened 2025-07-22
added example for return bytes

se7enarianelabs:
Closes: #66 

@jaycoolh We should probably add a helper function to the kit to retrieve the bytes in the next release of the agent kit

### DIFF
--- a/typescript/examples/langchain/package.json
+++ b/typescript/examples/langchain/package.json
@@ -4,7 +4,8 @@
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
     "langchain:structured-chat-agent": "ts-node structured-chat-agent.ts",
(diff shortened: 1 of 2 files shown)


typescript/examples/langchain/package.json
```diff
@@ -4,7 +4,8 @@
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
     "langchain:structured-chat-agent": "ts-node structured-chat-agent.ts",
-    "langchain:tool-calling-agent": "ts-node tool-calling-agent.ts"
+    "langchain:tool-calling-agent": "ts-node tool-calling-agent.ts",
+    "langchain:return-bytes-tool-calling-agent": "ts-node return-bytes-tool-calling-agent.ts"
```
jaycoolh, line 8: I think we can omit the `langchain:` bit as you have to be in the langchain directory to run this XD

PR #105, opened 2025-07-25
#92 tinybar -> hbar conversion for account balance

marcin-piela-ariane:
Closes: #92


jaycoolh (approved):

jaycoolh:
@marcin-piela-ariane how did you test this?

marcin-piela-ariane:
@jaycoolh

> how did you test this?

Using prompt. Without it the agent always returned false balance: 100,000,000,000 tinybars i.e. 100 hbar (according to AI which is false).

I also looked-up the raw query response.

Issue #102, opened 2025-07-24
Add Transfer ERC20 Tool


se7enarianelabs:
### Description:
We need to implement a new tool for transferring fungible tokens (ERC20-equivalent) on the Hedera network. This tool should allow users to transfer tokens from their account to another using the `trasfer` function. The tool should support single transfers, it should support both EVM and Hedera accountId addresses.

### Acceptance Criteria:
- [ ] New tool file created under src/shared/tools/erc20/ (e.g., transfer-erc20.ts)
- [ ] Tool accepts parameters: token address (string, required), amount (number, required), destination address (string, required, eg. 0.0.xxxxx or 0xEEEEEEEEEEE)
- [ ] Proper error handling and validation for input parameters
- [ ] Add Zod schema for input validation in parameter-schemas/erc20.zod.ts
- [ ] Add tool registration in src/shared/tools.ts and update toolkit(s) as needed

### Additional Notes:
Follow the structure and conventions of existing FT tools (e.g., create-fungible-token.ts, airdrop-fungible-token.ts)
Support both account IDs and account aliases for destination accounts. Consider adding examples in the tool description for common use cases Remember to test the tool.

walkerlj0:
@mgarbs to review

PR #115, opened 2025-07-31
#102 New tool added: transfering of ERC20 tokens

marcin-piela-ariane:


walkerlj0:
@mgarbs to review

typescript/src/shared/hedera-utils/hedera-parameter-normaliser.ts
```diff
@@ -273,4 +276,14 @@ export default class HederaParameterNormaliser {
...
     };
   }
+
+  static normaliseTransferERC20Params(
+    params: z.infer<ReturnType<typeof transferERC20Parameters>>,
+    _context: Context,
+  ) {
+    return {
+      ...params,
+      gas: 10_000_000,
```
jaycoolh, line 286: very high gas limit

typescript/src/shared/hedera-utils/hedera-builder.ts
```diff
@@ -64,4 +69,15 @@ export default class HederaBuilder {
...
+  static transferERC20(
+    params: z.infer<ReturnType<typeof transferERC20ParametersNormalised>>,
+  ) {
+    return new ContractExecuteTransaction()
+      .setContractId(params.contractId)
+      .setGas(params.gas)
+      .setFunction("transfer", new ContractFunctionParameters()
+        .addAddress(params.recipientAddress)
+        .addUint256(params.amount));
+  }
```
jaycoolh, line 82: This doesn't follow the approach we have taken in the past where we pass args in constructor

typescript/src/plugins/core-scs-plugin/tools/erc20/transfer-erc20.ts
```diff
@@ -0,0 +1,61 @@
...
+This tool will transfer a given amount of an existing ERC20 token on Hedera.
+
+Parameters:
+- contractId (str, required): The id of the ERC20 contract
+- recipientAddress (str, required): The EVM address to which the tokens will be transfered
+- amount (number, required): The amount to be transfered
+${usageInstructions}
+
+Example: "Transfer 1 ERC20 token 0.0.6473135 to 0xd94dc7f82f103757f715514e4a37186be6e4580b" means transfering the amount of 1 of the ERC20 token with contract id 0.0.6473135 to the 0xd94dc7f82f103757f715514e4a37186be6e4580b EVM address.
+`;
```
jaycoolh, line 27: It's unlikely that, if using the ERC-20 tool, the contract will be referred to as 0.0.xxx

Instead it will probably be the 0x address. @mgarbs have you reviewed this PR? - does the current implementation work with 0x addresses? Or do we need to update the `HederaParameterNormaliser.normaliseTransferERC20Params` so that we fetch the contract id from the evm addresses

jaycoolh (changes requested):

typescript/src/shared/hedera-utils/hedera-parameter-normaliser.ts
```diff
@@ -302,4 +305,56 @@ export default class HederaParameterNormaliser {
...
+    mirrorNode: IHederaMirrornodeService,
+  ) {
+    const recipientAddress = await HederaParameterNormaliser.getHederaEVMAddress(
+      params.recipientAddress,
+      mirrorNode,
+    );
+    const contractId = await HederaParameterNormaliser.getHederaAccountId(
+      params.contractId,
+      mirrorNode,
+    );
```
mgarbs, line 323: This uses `getHederaAccountId` for a contract address. While functionally correct, consider renaming to getContractId or getHederaAddressId for better readability and intent.

typescript/src/shared/hedera-utils/hedera-parameter-normaliser.ts
```diff
@@ -302,4 +305,56 @@ export default class HederaParameterNormaliser {
...
+  static async getHederaAccountId(
+    address: string,
+    mirrorNode: IHederaMirrornodeService,
+  ): Promise<string> {
+    if (AccountResolver.isHederaAddress(address)) {
+      return address;
+    }
+    const account = await mirrorNode.getAccount(address);
+    return account.accountId;
+  }
```
mgarbs, line 359: This method name suggests it's for accounts only, but it's also used for contracts. Consider renaming or creating separate methods for contracts vs accounts for clarity.

typescript/src/plugins/core-scs-plugin/tools/erc20/transfer-erc20.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    console.log('normalisedParams', JSON.stringify(normalisedParams, null, 2));
+    const tx = HederaBuilder.executeTransaction(normalisedParams);
+    const result = await handleTransaction(tx, client, context);
+    return result;
+  } catch (error) {
+    if (error instanceof Error) {
+      return error.message;
+    }
+    return 'Failed to transfer ERC20';
```
mgarbs, line 60: Is there a reason for returning the error instead of throwing it?

typescript/src/shared/hedera-utils/hedera-parameter-normaliser.ts
se7enarianelabs: from Hedera mirrornode point of view the name is correct, in the mirrornode returns the account id of a contract. Having the same function with diffrent names introduces complexity and can confuse developers

typescript/src/shared/hedera-utils/hedera-parameter-normaliser.ts
se7enarianelabs: from Hedera mirrornode point of view the name is correct, in the mirrornode returns the account id of a contract. Having the same function with diffrent names introduces complexity and can confuse developers

typescript/src/shared/hedera-utils/hedera-parameter-normaliser.ts
se7enarianelabs: @mgarbs 
The api does not distinct between contracts and accounts, at least on the endpoint that we are using:
- Account: https://testnet.mirrornode.hedera.com/api/v1/accounts/0.0.6360977
- Contract: https://testnet.mirrornode.hedera.com/api/v1/accounts/0.0.6551590

As you see in the example above, we are using the same endpoint for both entities

mgarbs: I get that. We do have 
GET [/api/v1/contracts/{contractIdOrAddress}](https://testnet.mirrornode.hedera.com/api/v1/docs/#/contracts/getContract)

If you feel so strongly that the code should not get a contract via the get contract endpoint and get an account via the get account endpoint then I am not going to get in your way.

PR #125, opened 2025-08-04
Add memejob plugin, toolkits abstractization

vla-dev:
**This PR includes the following:**
- memejob plugin that can create, buy, and sell memecoins
- toolkit abstractization, where each toolkit implementation is simplified by extending the `HederaBaseToolkit` class
- tsup tweaks to keep the `@hashgraph/sdk` external
- memejob plugin usage examples for both `ai-sdk` and `langchain` toolkits

**Reviewer note:**
When installing the memejob plugin from the local path (e.g. `file:../../src/plugins/memejob-plugin`), it's not hoisting the common dependencies  (`@hashgraph/sdk`) between the examples package and the rest of the packages, which means the local dependencies are not deduped. 

In order to test it properly, we recommend building the memejob-plugin package and packing it into a tarball, which can later be installed in the example package. Otherwise, the plugin must be published on npm and installed directly with the `latest` tag.


piotrswierzy (commented):
Plugin looks nice, as discuss please have it in your own repo for now. 
Regarding the `HederaBaseToolkit`, nice cleanup, but I don’t think inheritance buys us much here. 
We’re just mapping `Tool[] → HederaAgentKitTool[]` and wiring deps. That’s maybe composition territory, but not a lifecycle that needs a base class.

se7enarianelabs (commented):
Plugin looks nice, as discuss please have it in your own repo for now.
Regarding the `HederaBaseToolkit`, nice cleanup, but I don’t think inheritance buys us much here.
We’re just mapping `Tool[] → HederaAgentKitTool[]` and wiring deps. That’s maybe composition territory, but not a lifecycle that needs a base class.

Issue #133, opened 2025-08-06
Add Mint ERC721 tool


se7enarianelabs:
### Description:
We need to implement a new tool for minting ERC721 on the Hedera network. This tool should allow users to mint tokens to an account using the `safeMint` function. The tool should support single minting, it should support both EVM and Hedera accountId addresses.

### Acceptance Criteria:
- [ ] New tool file created under src/plugins/core-erc721-plugin/tools (e.g., transfer-erc20.ts)
- [ ] Tool accepts parameter: token address (string, required), destination address (string, required, eg. 0.0.xxxxx or 0xEEEEEEEEEEE)
- [ ] Proper error handling and validation for input parameters
- [ ] Add Zod schema for input validation in parameter-schemas/erc721.zod.ts
- [ ] Add tool to plugin

### Additional Notes:
Follow the structure and conventions of existing ERC721 tools
Support both account IDs and account aliases for destination accounts. Consider adding examples in the tool description for common use cases ,Remember to test the tool.

walkerlj0:
@skurzyp-arianelabs to review this one & if approved then merge

PR #144, opened 2025-08-11
Feat/mint erc721 tool

se7enarianelabs:
Closes: #133

### DIFF
--- a/typescript/src/plugins/core-scs-plugin/tools/erc721/mint-erc721.ts
+++ b/typescript/src/plugins/core-scs-plugin/tools/erc721/mint-erc721.ts
@@ -35,7 +35,6 @@ const mintERC721 = async (
   params: z.infer<ReturnType<typeof mintERC721Parameters>>,
 ) => {
   try {
-    console.log('mintERC721', JSON.stringify(params, null, 2));
     const mirrorNode = getMirrornodeService(context.mirrornodeService, client.ledgerId!);
 
     const normalisedParams = await HederaParameterNormaliser.normaliseMintERC721Params(
@@ -45,7 +44,6 @@ const mintERC721 = async (
       context,
       mirrorNode,
     );
-    console.log('normalisedParams', JSON.stringify(normalisedParams, null, 2));
     const tx = HederaBuilder.executeTransaction(normalisedParams);
     console.log('tx', JSON.stringify(tx, null, 2));
     const result = await handleTransaction(tx, client, context);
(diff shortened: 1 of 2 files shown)


typescript/src/plugins/core-scs-plugin/tools/erc721/mint-erc721.ts
```diff
@@ -0,0 +1,71 @@
...
+    const normalisedParams = await HederaParameterNormaliser.normaliseMintERC721Params(
+      params,
+      ERC721_MINT_FUNCTION_ABI,
+      ERC721_MINT_FUNCTION_NAME,
+      context,
+      mirrorNode,
+    );
+    console.log('normalisedParams', JSON.stringify(normalisedParams, null, 2));
+    const tx = HederaBuilder.executeTransaction(normalisedParams);
+    console.log('tx', JSON.stringify(tx, null, 2));
```
skurzyp-arianelabs, line 50: Do we want those `console.log`s to stay? If so, add the same to `transfer-erc20`.

typescript/src/plugins/core-scs-plugin/tools/erc721/transfer-erc721.ts
```diff
@@ -0,0 +1,76 @@
...
+      params,
+      ERC721_TRANSFER_FUNCTION_ABI,
+      ERC721_TRANSFER_FUNCTION_NAME,
+      context,
+      mirrorNode,
+    );
+    console.log('normalisedParams', JSON.stringify(normalisedParams, null, 2));
+    const tx = HederaBuilder.executeTransaction(normalisedParams);
+    console.log('tx', JSON.stringify(tx, null, 2));
+    const result = await handleTransaction(tx, client, context);
```
skurzyp-arianelabs, line 56: Same as above - should those `console.log`s stay? Do we need 3 of them?

typescript/src/shared/parameter-schemas/hscs.zod.ts
```diff
@@ -9,3 +9,38 @@ export const contractExecuteTransactionParametersNormalised = (_context: Context
...
+    tokenName: z.string().describe('The name of the token.'),
+    tokenSymbol: z.string().describe('The symbol of the token.'),
+    decimals: z
+      .number()
+      .int()
+      .min(0)
+      .max(18)
+      .default(18)
+      .describe('The number of decimals the token supports.'),
+    initialSupply: z.number().int().min(0).default(0).describe('The initial supply of the token.'),
```
skurzyp-arianelabs, line 31: if `initialSupply` is not optional then the default value is not necessary - mby consider making it optional like in case of Hedera FTs?

typescript/src/plugins/core-scs-plugin/tools/erc721/transfer-erc721.ts
se7enarianelabs: nope, should be removed

PR #146, opened 2025-08-12
fix: fix the missing default max supply for token creation

### LINKED ISSUE #145: Bug: Missing Default `maxSupply` in tools params normalization for Token Creation Causes Errors in ElizaOS plugin
# Bug: Missing Default `maxSupply` in tools params normalization for Token Creation Causes Errors in ElizaOS plugin

## Description

When using ElizaOS to create a token with the prompt:

```
Create token Gold with symbol GLD, 3 decimals and finite supply type
```

the agent fails during **parameter normalization**.

This doesn't happen in LangChain because LangChain's tool description includes a default for `maxSupply`:

```
- supplyType (str, optional): The supply type of the token. Can be "finite" or "infinite". Defaults to "finite"
- maxSupply (int, optional): The maximum supply of the token. Only applicable if supplyType is "finite". Defaults to 1,000,000 if not specified
```

ElizaOS builds action descriptions from the Zod schema which currently does not encode a `maxSupply` default. As a result `maxSupply` stays `undefined`, the normalization logic can't handle this case and throws an error. Despite the error, ElizaOS still attempts to execute the action and produces hallucinated results.

---

## Steps to Reproduce

1. Start ElizaOS and the Hedera plugin MVP following the README: [https://github.com/skurzyp-arianelabs/elizaos-hedera-plugin-mvp](https://github.com/skurzyp-arianelabs/elizaos-hedera-plugin-mvp)
2. Prompt the agent with:

```
Create token Gold with symbol GLD, 3 decimals and finite supply type
```

3. Observe CLI logs. An error occurs during parameter normalization and then the action is (incorrectly) executed.

---

## Expected Behavior

* If `supplyType` is `"finite"` and `maxSupply` is not provided, the system should set `maxSupply` to the default value `1_000_000` before normalization.
* No error should be thrown and no hallucinated result should occur.

---

## Actual Behavior

* `maxSupply` remains `undefined` because the Zod schema does not provide a default.
* Parameter normalization fails and throws an error.
* ElizaOS proceeds to call the action and returns fabricated results.

## Additional Notes

This is a schema/description mismatch: LangChain’s tool descriptions bake in defaults at the tool layer while ElizaOS builds descriptions from ZodSchemas that currently do not carry the same defaults. Aligning them (schema defaults or normalization fallbacks) will fix the error and prevent hallucinated executions.

skurzyp-arianelabs:
Fixes #145.


se7enarianelabs (dismissed):
LGTM!

se7enarianelabs (approved):

Issue #168, opened 2025-08-13
feat: core-queries-plugin: TOKEN_INFO_QUERY_TOOL


walkerlj0:
Plugin: core-queries-plugin
Project Tag: 3.2.0

## User Story
As a developer, I want TOKEN_INFO_QUERY_TOOL to fetch token metadata and settings for a given token ID.

## Files Changed
- [ ] typescript/src/plugins/core-queries-plugin/tools/token-info-query.ts (new)
- [ ] typescript/src/plugins/core-queries-plugin/index.ts (import, register, export tool)
- [ ] docs/HEDERAPLUGINS.md (add tool to Queries section)

## Acceptance Criteria
- Uses TokenInfoQuery with input: tokenId (required).
- Handles errors gracefully; clear messages for invalid token ID or network issues.
- Returns structured result with key fields (name, symbol, tokenType, decimals, totalSupply, treasuryAccountId, kyc/freeze flags, admin/supply/kyc/freeze keys, autoRenew info, memo).

walkerlj0:
@se7enarianelabs will check and merge if done

PR #174, opened 2025-08-14
feat: add token info query action

skurzyp-arianelabs:
This PR references #168.
Changes:
- new action added - `CREATE_FUNGIBLE_TOKEN_TOOL`
- directory within `core-queries-plugin` renamed from `account-queries` to just `queries` as currently in addition to account related tools it contains topic and token query tools


skurzyp-arianelabs:
@se7enarianelabs please review it. I've been developing this tool couple days ago before we had a talk about the naming of plugins/tools and I tried to align some things. Based on today's discussion i know that probably those changes should be reverted and some bigger refactor will be performed soon.

PR #187, opened 2025-08-19
fix: elisaos adapter and dependencies

skurzyp-arianelabs:
**Description**:
This PR resolves the `ERR_PACKAGE_PATH_NOT_EXPORTED` error occurring when running LangChain example scripts that depend on @elizaos/core.
The issue was caused by @elizaos/core lacking a support for CommonJS.
This PR also adds previously deleted `elizaos-adapter`.

**Related issue(s)**:

#183 , #122 

**Notes for reviewer**:
This changes nothing from the developer perspective. The API, library installation, etc are still the same.
I have decided to switch to `tsx` in langchain and leave the `ts-node` in the `ai-sdk` to highlight different possibilities. In addition to that, `tsx` is newer and more flexible solution.

**Tested by running:**
- langchain example (CommonJS flow)
- elizaOS plugin (ESM flow)

**Checklist**

- [X] Documented (Code comments, README, etc.)
- [X] Tested (unit, integration, etc.)


se7enarianelabs (commented):
lgtm

se7enarianelabs (approved):

typescript/src/elizaos/adapter.ts
jaycoolh, line 1: I personally think it would be nicer to follow the same conventions we have for our other adapters where we have

`elizaos/toolkit.ts`
`elizaos/tool.ts`

Where tool.ts contains a class that instantiates the `Action` object and we map over the tools in the constructor. then we calling `.getActions` it returns the array from memory rather than creating a new array (same pattern as we have in the other adapters)

I know eliza calls them actions and not tools but they are synonyms and _I think_ for devs who are looking at the source code, it will make more sense to it following the same structure as others.

thoughts? @skurzyp-arianelabs @se7enarianelabs

jaycoolh: Even if we don't call them `tools` internally, I think creating the list of `Actions` in the constructor and separating `Action` creation from the Adapter class, so it confirms with other code, is sensible

typescript/src/elizaos/index.ts
jaycoolh, line 1: Do we have an example for this? We should

typescript/README.md
```diff
@@ -56,10 +56,10 @@ npm init -y
 npm install hedera-agent-kit @langchain/openai @langchain/core langchain @hashgraph/sdk dotenv
 ```
 
-> You may need to add "type": "module" to your package.json file so that your project allows ES module imports
+> Note: The examples in the `typescript/examples` directory are configured for `"type": "commonjs"`, whereas the quick-start guide below demonstrates an ES module setup. For more information on ES module and CommonJS interoperability, please see the [TypeScript documentation](https://www.typescriptlang.org/docs/handbook/modules/appendices/esm-cjs-interop.html).
```
jaycoolh, line 59: I'm not sure if we need this... best case scenario it will distract user from our docs and maybe make them go elsewhere, worse case it'll confuse them

jaycoolh (changes requested):

typescript/src/elizaos/adapter.ts
skurzyp-arianelabs: Sounds good to me :+1:

typescript/src/elizaos/index.ts
skurzyp-arianelabs: Should I add readme there with example of usage and link to our `elizaos-hedera-plugin` repo? Currently the plugin code there is not merged to main yet but soon will be. O maybe the example should be added to the main `README.MD` file?

typescript/src/elizaos/adapter.ts
se7enarianelabs: Good point

typescript/src/elizaos/index.ts
jaycoolh: Does the `elizaos-hedera-plugin` repo have an example implementation?

typescript/src/elizaos/tool.ts
```diff
@@ -0,0 +1,123 @@
...
+            text: `Execution failed: ${message}`,
+            content: { error: message },
+          });
+        }
+
+        return { success: false, text: `Execution failed: ${message}`, error: message };
+      }
+    };
+  }
+  handler: Handler;
```
jaycoolh, line 116: can we put this above the constructor XD

jaycoolh (changes requested):
Looking good. Also need to add:

1. docs / readme pointing to the main hedera eliza plugin and reference to the example implementation
2. Add an ADR to document bundling decision (ref https://github.com/hashgraph/hedera-agent-kit/issues/122#issuecomment-3196871939)

typescript/src/elizaos/tool.ts
skurzyp-arianelabs: Done

jaycoolh (dismissed):

PR #292, opened 2025-09-16
fix: typing issues with typescript and zod schemas

skurzyp-arianelabs:
**Description**:
This PR fixes TS error that do not affect how the code works.
 Errors in `create-account.integration.test.ts` are caused by TS not knowing that some of the object properties are optional and have default values even tho the are defined like that in the zod schemas. 
 Errors in Hedera parameter normalizer are caused by narrowing the return type of function despite returning more data.

**Checklist**
- [X] Documented (Code comments, README, etc.)
- [X] Tested (unit, integration, etc.)

### DIFF
--- a/typescript/src/shared/hedera-utils/hedera-parameter-normaliser.ts
+++ b/typescript/src/shared/hedera-utils/hedera-parameter-normaliser.ts
@@ -35,7 +35,6 @@ import { TokenTransferMinimalParams, TransferHbarInput } from '@/shared/hedera-u
 import { AccountResolver } from '@/shared/utils/account-resolver';
 import { ethers } from 'ethers';
 import {
-  contractExecuteTransactionParametersNormalised,
   createERC20Parameters,
   createERC721Parameters,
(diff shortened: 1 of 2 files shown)


typescript/src/shared/hedera-utils/hedera-parameter-normaliser.ts
```diff
@@ -342,7 +341,7 @@ export default class HederaParameterNormaliser {
     factoryContractAbi: string[],
     factoryContractFunctionName: string,
     context: Context,
-  ): z.infer<ReturnType<typeof contractExecuteTransactionParametersNormalised>> {
```
mmalik-al, line 345: @skurzyp-arianelabs Wouldn’t it be better to modify the return type instead of just removing it? contractExecuteTransactionParametersNormalised is missing a few fields, but you can combine it with createERC20Parameters, and then it should work fine. 😃

mmalik-al (approved):

PR #673, opened 2026-04-10
refactor: migrate transaction bytes handling to Uint8Array (#672)

### LINKED ISSUE #490: migration to uInt8Array from Buffer in RETURN_BYTES mode
### Issue: Inconsistent Transaction Bytes Serialization Across Environments

We currently have special-case logic for extracting transaction bytes due to differences between the frontend (web) and backend (Node.js) runtimes.

In the Node.js environment, byte data is returned as a `Buffer`. However, the web runtime does not support `Buffer`, so we rely on polyfills. These polyfills alter how the byte data is serialized, resulting in two different representations of the same underlying data.

#### Node.js output

```json
{
  "toolName": "create_topic_tool",
  "toolCallId": "call_OMa393aZClrGCbQNk1BoIDaJ",
  "parsedData": {
    "raw": {
      "bytes": {
        "type": "Buffer",
        "data": [
          10,
          81,
          42,
          79,
          10
        ]
      }
    }
  }
}
```

#### Web engine output

```json
{
  "raw": {
    "bytes": {
      "0": 10,
      "1": 75,
      "2": 42,
      "3": 73,
      "4": 10,
      "5": 69,
      "6": 10,
      "7": 26,
      "8": 10,
      "9": 11,
      "10": 8,
      "11": 138,
      "12": 239,
      "13": 221,
      "14": 203
    }
  }
}
```

Because of this discrepancy, the current implementation includes custom logic to detect and normalize both formats after byte extraction. While functional, this approach increases complexity and maintenance overhead.

#### Proposed Improvement

A cleaner and more robust solution would be to standardize the internal byte representation within HAK to use `Uint8Array`. `Uint8Array` is natively supported in both Node.js and web environments and would eliminate the need for polyfills and special-case handling.

#### Impact

Adopting `Uint8Array` as the canonical byte format would be a **breaking change**, as it alters the internal and potentially external data contract. This change would require:

* Clear migration guidance
* Versioning considerations
* Updates to any consumers relying on the current `Buffer`-based structure

MWBlocky:
**Related issue(s)**:

Fixes #490


skurzyp-blockydevs (dismissed):

docs/DEVEXAMPLES.md
```diff
@@ -147,18 +147,24 @@ npm install
...
+  : Buffer.from(bytesObject.data);
 ```
 
-**LangChain v1:**
-```bash
-npm run langchain:return-bytes-tool-calling-agent-web
+After:
+
+```ts
+const realBytes = Uint8Array.from(bytesObject);
```
skurzyp-blockydevs, line 165: i would suggest not using the `bytesObject` name anymore. Also, it seems that we don't even need to cast to the `Uint8Array`. We can just extract the `bytes` from the payload and pass them to the `Transaction.fromBytes(bytes)` and it worked out of box in my case. Casting is optional and only required for type checking.

typescript/examples/langchain-v1/README.md
```diff
@@ -28,13 +28,23 @@ An agent that returns transaction bytes for **manual signing and execution**, ra
...
+const realBytes = Buffer.isBuffer(bytesObject)
+  ? bytesObject
+  : Buffer.from(bytesObject.data);
 ```
 
-A variant of the Return Bytes agent with **robust parsing logic** for handling multiple Buffer serialization formats (for example, browser `Uint8Array` and JSON-serialized buffers).
+After:
+
+```ts
+const realBytes = Uint8Array.from(bytesObject);
```
skurzyp-blockydevs: same as above

typescript/examples/langchain-v1/return-bytes-tool-calling-agent.ts
```diff
@@ -86,9 +86,7 @@ async function bootstrap(): Promise<void> {
       if (toolCall.parsedData?.raw?.bytes) {
         console.log('Transaction bytes found. Executing...');
         const bytesObject = toolCall.parsedData.raw.bytes;
-        const realBytes = Buffer.isBuffer(bytesObject)
-          ? bytesObject
-          : Buffer.from(bytesObject.data);
+        const realBytes = Uint8Array.from(bytesObject);
 
         const tx = Transaction.fromBytes(realBytes);
```
skurzyp-blockydevs: like above:
```
const bytes = toolCall.parsedData.raw.bytes;
const tx = Transaction.fromBytes(bytes);
```
worked in my case - the `bytesObject` and `realBytes` var names might be confusing. For type clarity we can inclued the `Uint8Array.from()` step but I'm not sure if it shouldnt be more like `bytes as Uint8Array`. What do you think?

typescript/examples/langchain/return-bytes-tool-calling-agent.ts
```diff
@@ -94,7 +94,7 @@ async function bootstrap(): Promise<void> {
       console.log(`AI: ${response?.output ?? response}`);
       const bytes = extractBytesFromAgentResponse(response);
       if (bytes !== undefined) {
-        const realBytes = Buffer.isBuffer(bytes) ? bytes : Buffer.from(bytes.data);
+        const realBytes = Uint8Array.from(bytes);
         const tx = Transaction.fromBytes(realBytes);
```
skurzyp-blockydevs: this is closer to what I would expect to see as a developer - still not sure about the ` Uint8Array.from(bytes)` step tho

typescript/src/shared/strategies/tx-mode-strategy.ts
```diff
@@ -65,7 +65,7 @@ class ReturnBytesStrategy implements TxModeStrategy {
     if (!context.accountId) throw new Error('…');
     const id = TransactionId.generate(context.accountId);
     tx.setTransactionId(id).freezeWith(client);
-    return { bytes: tx.toBytes() };
+    return { bytes: Array.from(tx.toBytes()) };
   }
 }
```
skurzyp-blockydevs, line 70: Do you think that adding unit tests testing it should be in scope of this PR? It feels like it is a good moment to do that.

skurzyp-blockydevs (changes requested):

docs/DEVEXAMPLES.md
MWBlocky: fixed in b13a87bb80bc4776b65f61c51726e238b13bc16b

typescript/examples/langchain-v1/README.md
MWBlocky: fixed in b13a87bb80bc4776b65f61c51726e238b13bc16b

typescript/examples/langchain-v1/return-bytes-tool-calling-agent.ts
MWBlocky: fixed in b13a87bb80bc4776b65f61c51726e238b13bc16b

typescript/examples/langchain/return-bytes-tool-calling-agent.ts
MWBlocky: fixed in b13a87bb80bc4776b65f61c51726e238b13bc16b

skurzyp-blockydevs (approved):
Please change the target branch to `feat/release/16.04` and resolve the conflicts. Then its good to be merged  😄

PR #699, opened 2026-04-14
chore(docs): update docs to reflect MCP changes

kierzniak:


MWBlocky (approved):

examples/modelcontextprotocol/README.md
```diff
@@ -58,17 +58,18 @@ HEDERA_OPERATOR_KEY=302e...
...
-pnpm start:http
+npm run start:http:return-bytes
 ```
 
-### HTTP Server (Return Bytes Mode)
+To run the HTTP server in autonomous mode, invoke `src/http.ts` directly:
+
 ```bash
-pnpm start:http:return-bytes
+npx tsx src/http.ts
```
skurzyp-blockydevs: please remove this part about runnig the HTTP server in autonomus mode. Current implementation does not support it - it is just for the return bytes mode. Of course it could be modified if needed it is not ready right now.

docs/MIGRATION-v4.md
```diff
@@ -429,41 +424,51 @@ const toolkit = new HederaElizaOSToolkit({
...
-        "HEDERA_OPERATOR_ID": "0.0.xxxxx",
-        "HEDERA_OPERATOR_KEY": "302e..."
-      }
-    }
-  }
-}
+const server = new HederaMCPToolkit({
+  client,
+  configuration: {
+    plugins: [coreTokenPlugin, coreAccountPlugin],
```
skurzyp-blockydevs, line 461: maybe we should add a note that now en empty array will not resolve to import of all tools?

skurzyp-blockydevs (changes requested):

docs/MIGRATION-v4.md
kierzniak: We mention this here: https://github.com/hashgraph/hedera-agent-kit-js/blob/feat/release/16.04-docs/docs/MIGRATION-v4.md#4-explicit-plugin-opt-in-behavioral-change

examples/modelcontextprotocol/README.md
kierzniak: Removed

skurzyp-blockydevs (approved):

PR #742, opened 2026-04-27
feat: add solo as default test network

kierzniak:
**What changed**

Before this branch, network-specific decisions were scattered across at least six files. client-setup.ts decided which SDK client to build, `test-balance-tiers.ts` decided how much HBAR to fund, `account-teardown.ts` decided how to clean up, `usd-to-hbar-service.ts` decided how to price funding, and so on. Each one independently called `getTestNetwork()` and branched on `'local-node'` vs `'testnet'`, which meant adding Solo support touched every one of them and the next network would touch them all again. Tests themselves also had to know which network they were on to pick the right helper.

The new TestProfile abstraction in `packages/tests/shared/profile/` collapses that branching into two adapters (`soloProfile`, `testnetProfile`) behind one interface, with `getProfile()` resolving the active one from `HEDERA_NETWORK`. The interface is small and intentionally shaped around what tests actually need: `profile.accounts.acquire({ tier })` and `release()` for funded test identities, `profile.client.connectAs(account)` for an SDK client plus ops wrapper, `profile.balance.fund(tier)` and `usdToHbar(usd)` for funding amounts that adapt to the network's exchange rate, `profile.operator` for the env-derived operator account, and `profile.dispose()` for session cleanup. Tests no longer reference `getTestNetwork()` at all, and ~91 test files were migrated to this API. Adding a new network in the future means writing one more adapter, not editing six files.

Alongside the migration, the test infra itself got cleaned up. Fixed `wait(MIRROR_NODE_WAITING_TIME)` sleeps were replaced with adaptive `waitForMirrorTx` polling, the duplicate `itWithRetry` retry layer was removed in favor of vitest's built-in retry: 3, and `verifyHbarBalanceChange` was tightened from "balance didn't go down" to strict equality with documented fee semantics.

To support local Solo workflows, `scripts/solo/` now contains lifecycle scripts (`install`, `deploy`, `check`, `deploy:contracts`, `destroy`) wired into root pnpm scripts, with up and down aggregates. CI and local share the same commands. The readiness gate (`scripts/solo/check.sh`) probes consensus gRPC, mirror REST, the mirror Web3 simulator, and the JSON-RPC relay; the Web3 probe catches the simulator warm-up window that used to break contract deploys with 503s. CI was also parallelized so integration and e2e jobs run after unit tests in parallel, with `maxWorkers: 8` per vitest run.

**Impact**

Local development is now self-contained. There's no need for testnet credentials or rate-limit budgets; pnpm test:solo:up brings up a working network in about a 3 - 5 minutes. Per-test latency dropped roughly 4 to 5x, with Solo runs averaging 15 to 25 seconds per test vs 60 to 95 on testnet. Test code is cleaner because there are no if (network === ...) branches, fee accounting is explicit instead of hidden behind a loose assertion, and there's only one retry layer to reason about. Because CI and local use the same commands, reproducing CI failures locally is straightforward.

**Where to look**

The full local workflow, prerequisites, and troubleshooting live in docs/TESTING.md. ADR document docs/adr/004-dynamic-test-account-balances.md was changed to reflect the new changes.


MWBlocky:
# Test Execution Summary

I followed the steps described in `docs/TESTING.md` and successfully executed the full test suite (unit, integration, and e2e). All test stages completed successfully.

---

## Results Overview

## Unit Tests

| Type | Package | Passed | Skipped | Duration |
|------|--------|--------|----------|----------|
| Unit | core | 502 | 0 | 7.74s |
| Unit | ai-sdk | 0 (no tests) | 0 | 0.02s |
| Unit | elizaos | 0 (no tests) | 0 | 0.03s |
| Unit | langchain | 0 (no tests) | 0 | 0.02s |

---

## Integration Tests

| Type | Package | Passed | Skipped | Duration |
|------|--------|--------|----------|----------|
| Integration | core | 206 | 2 | 57.47s |
| Integration | langchain | 123 | 89 | 174.57s |
| Integration | ai-sdk | 0 (no tests) | 0 | 0.02s |
| Integration | elizaos | 0 (no tests) | 0 | 0.02s |

---

## E2E Tests

| Type | Package | Passed | Skipped | Duration |
|------|--------|--------|----------|----------|
| E2E | langchain | 137 | 2 | 320.39s |
| E2E | core | 0 | 0 | 0.02s |
| E2E | ai-sdk | 0 | 0 | 0.02s |
| E2E | elizaos | 0 | 0 | 0.02s |
| E2E | tests package | — | — | moved to langchain |

  What each one covers:

  - core (@hashgraph/hedera-agent-kit) — the framework-agnostic engine: tool definitions, plugins, transaction builders, Hedera SDK wiring. Most logic lives here, which is why it carries the bulk of unit tests (502) and a large integration suite (206) hitting the network.
  - langchain — the LangChain adapter that wraps core's tools as LangChain Tools. It has integration + e2e tests because the real validation is "does the agent loop, with an LLM, actually call these tools end-to-end?" That's why the e2e tests landed here.
  - ai-sdk — Vercel AI SDK adapter. Same shape as langchain, but currently no tests written yet — the test:unit script is literally an echo 'No tests…' placeholder.
  - elizaos — ElizaOS adapter, same situation: scaffolded package, no tests yet.

---

All skipped tests in the suite, by package. (Reported counts from vitest: core int 2, langchain int 89, langchain e2e 2 = 93.)

  core — integration (2)

  packages/core/tests/integration/hedera/transfer-fungible-token-with-allowance.integration.test.ts
  - L103 — should allow spender to transfer tokens to themselves using allowance
  - L135 — should allow spender to transfer tokens to themselves and receiver using allowance

  langchain — integration (89)

  All under packages/langchain/tests/integration/tool-matching/. Two patterns:

  Entire file skipped (describe.skip at the top)

  - delete-topic-tool-matching.integration.test.ts
  - update-token-tool-matching.integration.test.ts
  - get-hbar-balance-tool-matching.integration.test.ts
  - get-account-token-balances-tool-matching.integraion.test.ts (filename typo: "integraion")
  - get-account-query-tool-matching.integration.test.ts
  - get-token-info-query-tool-matching.integration.test.ts
  - get-contract-info-query-tool-matching.integration.test.ts
  - get-transaction-record-tool-matching.integration.test.ts
  - get-topic-messages-tool-matching.integration.test.ts
  - get-exchange-rate-tool-matching.integration.test.ts
  - dissociate-token-tool-matching.integration.test.ts
  - delete-account-tool-matching.integration.test.ts
  - schedule-delete-tool-matching.integration.test.ts
  - sign-schedule-transaction-tool-matching.integration.test.ts

  Only sub-suites skipped (rest of the file runs)

  - create-erc20-tool-matching.integration.test.ts — Tool Matching and Parameter Extraction + Tool Available
  - create-erc721-tool-matching.integration.test.ts — Tool Matching and Parameter Extraction + Tool Available
  - transfer-erc721-tool-matching.integration.test.ts — Tool Matching and Parameter Extraction + Tool Available
  - transfer-erc20-tool-matching.integration.test.ts — Tool Available
  - transfer-hbar-tool-matching.integration.test.ts — Tool Available
  - mint-fungible-token-tool-matching.integration.test.ts — Tool Available
  - mint-non-fungible-token-tool-matching.integration.test.ts — Tool Available
  - mint-erc721-tool-matching.integration.test.ts — Tool Available
  - airdrop-fungible-token-tool-matching.integration.test.ts — Tool Availability
  - get-topic-info-tool-matching.integration.test.ts — Tool Available

  langchain — e2e (2)

  - tests/e2e/create-account.e2e.test.ts:172 — should handle long memos correctly (comment: "LLM hallucinates that 60 is more than 100 and fails to even call the tool")
  - tests/e2e/mint-erc721.e2e.test.ts:107 — handles various natural language variations for minting


## Review observations
Looking at the skipped tests, almost all of them in `langchain` are tool-matching tests — the kind that check whether the LLM picks the right tool and extracts the right parameters from a natural-language prompt (89 integration + 2 e2e).

A few things I'm not sure about:

1. The `describe.skip` blocks turn off whole tool areas (delete-topic, update-token, sign-schedule-transaction, all the get-*-query matchers, etc.).  
   Is this meant to be temporary, or are these intentionally disabled for now?  
   If it's temporary, would it make sense to link them to an issue or TODO so they don't get lost over time?

2. The two e2e skips look like LLM misbehavior (one comment literally says the model thinks 60 > 100).  
   Are these considered known flakiness cases, or is the plan to fix prompts / constraints and re-enable them later?

3. One thing that surprised me: `core/.../transfer-fungible-token-with-allowance.integration.test.ts` also skips 2 cases (spender transferring to themselves using allowance).  
   These don't look LLM-related — is this a known SDK/network limitation, or just missing coverage?

Also a small idea — since the LLM tests are inherently probabilistic, would it make sense to keep them runnable but exclude them from the default `test:integration` run? For example a separate `test:llm` script. That way they wouldn’t get lost in skips, but also wouldn’t block CI due to model variability. Just a suggestion, not sure if it fits the intended test strategy.

packages/core/src/plugins/core-evm-plugin/tools/erc20/create-erc20.ts
```diff
@@ -63,7 +63,7 @@ export class CreateErc20Tool extends BaseTool {
     context: Context,
     client: Client,
   ) {
-    const factoryAddress = getERC20FactoryAddress(client.ledgerId!);
+    const factoryAddress = getERC20FactoryAddress();
```
skurzyp-blockydevs: why are we removing the param? I know that we dont have mainnet deployments but still passing a param for which ledger (mainnet | testnet | previewnet) do we want to get the addresses makes sense to me. Or does the `getERC20FactoryAddress` have access to the client instance and can handle this underneath?

packages/core/src/shared/constants/contracts.ts
```diff
@@ -37,29 +22,29 @@ export const ERC721_MINT_FUNCTION_NAME = 'safeMint';
...
+export function getERC20FactoryAddress(): string {
+  const address = process.env.HEDERA_ERC20_FACTORY_ADDRESS;
   if (!address) {
-    throw new Error(`Network type ${ledgerId} not supported for ERC20 factory`);
+    throw new Error(
+      'HEDERA_ERC20_FACTORY_ADDRESS is not set. See packages/core-contracts/README.md for the deployed testnet address, or run `pnpm test:solo:deploy:contracts` for Solo.',
+    );
   }
   return address;
 }
```
skurzyp-blockydevs: Are we moving the address to the env file now? So each user must configure one by themself? Isnt this a breaking change? This would mean that this PR introduces also changes to the core logic and we will have to release it.

disclaimer: i like the idea of having felxibility in terms of the contract address - i think this is a good UX if user knows what he's doing. I'm just worried about the breaking change.

.env.test.local.example
```diff
@@ -1,5 +1,60 @@
...
+# | testnet    | Your account ID | Your ECDSA key                                                       |
+# +------------+-----------------+----------------------------------------------------------------------+
+HEDERA_NETWORK=local-node
+HEDERA_ACCOUNT_ID=0.0.1002 # Solo default account; CI sets this automatically
+HEDERA_PRIVATE_KEY=0x105d050185ccb907fba04dd92d8de9e32c18305e097ab41dadda21489a211524 # Solo default ECDSA key; required by pnpm deploy:solo
+
+### EVM, contract tests
+# HEDERA_ECDSA_PRIVATE_KEY - ECDSA private key for the EVM signer that deploys / calls the factories
+# HEDERA_ERC20_FACTORY_ADDRESS - Address of the deployed ERC20 factory contract
+# HEDERA_ERC721_FACTORY_ADDRESS - Address of the deployed ERC721 factory contract
```
skurzyp-blockydevs: Do we need to set them now for normal usage too? If so, then please update the .env.example.

PS: see the comment above about the potential impact of this change

skurzyp-blockydevs (changes requested):

packages/core/src/plugins/core-evm-plugin/tools/erc20/create-erc20.ts
kierzniak: The whole codebase is already network-agnostic (network/operator/keys come from env), so the factory address resolution should follow the same pattern instead of being the one place that branches on ledgerId. The function reads HEDERA_ERC20_FACTORY_ADDRESS (and the ERC721 equivalent) directly. Users who want to point at a different network just set a different value. No need to thread client.ledgerId through anymore.

packages/core/src/shared/constants/contracts.ts
kierzniak: To avoid the breaking change, I'll add an if statement that falls back to the previous hardcoded testnet address when the env var isn't set on testnet, and display a deprecation notice pointing users at HEDERA_ERC20_FACTORY_ADDRESS (and the ERC721 equivalent). Existing testnet users keep working without config changes.

We can discuss how to improve this. Maybe we should add another tool to deploy/check contracts on different networks and remove the contract specific env vars?

.env.test.local.example
kierzniak: Yes, anyone using the ERC20/ERC721 tooling at runtime needs HEDERA_ERC20_FACTORY_ADDRESS / HEDERA_ERC721_FACTORY_ADDRESS. HBAR / HTS / query tooling is unaffected. I'll update the runtime .env.example files (examples/langchain/.env.example, etc.).

We can discuss how to improve this. Maybe we should add another tool to deploy/check contracts on different networks and remove the contract specific env vars?

packages/core/src/shared/constants/contracts.ts
skurzyp-blockydevs: Okey, I'm good with this change. We just need to make sure that:
1. the required envs are included in the `.env.example`
2. the required envs are mentioned in the `docs/HEDERATOOLS`, `docs/HEDERAPLUGINS`
3. mby we should extend the migration guide `docs/MIGRATION-v4`
4. mby we should add some banner in the readme? Like `> [!IMPORTANT] ...` 
5. it should be mentioned in the hedera docs: https://docs.hedera.com/hedera/open-source-solutions/ai-studio-on-hedera/hedera-ai-agent-kit

skurzyp-blockydevs: And we need to update the examples to have those env vars set.

kierzniak: I reverted these changes because I'm not sure about solution. This should be done in sparated EVM related PR.

.env.test.local.example
kierzniak: I reverted these changes because I'm not sure about solution. This should be done in sparated EVM related PR.

packages/core/src/plugins/core-evm-plugin/tools/erc20/create-erc20.ts
kierzniak: I reverted these changes because I'm not sure about solution. This should be done in sparated EVM related PR.

skurzyp-blockydevs (dismissed):

MWBlocky (dismissed):

skurzyp-blockydevs (dismissed):

.github/workflows/300-flow-pr-tests.yaml
```diff
@@ -0,0 +1,24 @@
+name: "300: [FLOW] PR Tests"
+
+on:
+  pull_request:
+    branches: [main]
```
andrewb1269:
```suggestion
    branches:
      - main
```

.github/workflows/300-flow-pr-tests.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  unit-tests:
+    uses: ./.github/workflows/800-call-run-unit-tests.yaml
+
+  integration-tests:
+    needs: [unit-tests]
```
andrewb1269:
```suggestion
    needs: 
      - unit-tests
```

.github/workflows/300-flow-pr-tests.yaml
```diff
@@ -0,0 +1,24 @@
...
+    uses: ./.github/workflows/800-call-run-integration-tests.yaml
+    secrets:
+      OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+
+  e2e-tests:
+    needs: [integration-tests]
```
andrewb1269:
```suggestion
    needs: 
      - integration-tests
```

.github/workflows/800-call-run-integration-tests.yaml
```diff
@@ -0,0 +1,93 @@
...
+    inputs:
+      node-version:
+        description: Node.js version
+        required: false
+        default: '22'
+        type: string
+    secrets:
+      OPENAI_API_KEY:
+        required: true
+    outputs: {}
```
andrewb1269:
```suggestion
```

andrewb1269 (changes requested):

.github/workflows/300-flow-pr-tests.yaml
kierzniak: Done!

.github/workflows/300-flow-pr-tests.yaml
kierzniak: Done!

.github/workflows/800-call-run-integration-tests.yaml
kierzniak: Done!

.github/workflows/300-flow-pr-tests.yaml
kierzniak: Done!

andrewb1269 (approved):

MWBlocky (approved):

PR #819, opened 2026-05-15
feat(create-hedera-agent): chat interface for the Hedera Agent Kit

### LINKED ISSUE #769: Create new version of web chat for HAK
Our current HAK web chat is outdated and no longer provides the level of user experience we want. It should be refreshed with a modern UI and improved usability to better match current standards. 

We can leverage the Chat UI project from Vercel as a foundation. This will allow us to quickly deliver a cleaner, more responsive interface with a better overall user experience.

The new chat UI should be included in the AgentLab download ZIP, along with documentation explaining how to deploy it to Vercel, including setup steps, required environment variables, and integration with the backend.

kierzniak:
Resolve #769


packages/create-hedera-agent/template/shared/config.js
```diff
@@ -0,0 +1,118 @@
...
+}
+
+function parseOperatorKey(key) {
+  const trimmed = key.trim();
+  if (/^303002/i.test(trimmed) || /^(0x)?[0-9a-fA-F]{64}$/.test(trimmed)) {
+    return PrivateKey.fromStringECDSA(trimmed);
+  }
+  return PrivateKey.fromStringED25519(trimmed);
+}
+
```
skurzyp-blockydevs: have you tested it with both types of keys?

packages/create-hedera-agent/template/web/src/app/layout.jsx
```diff
@@ -0,0 +1,41 @@
...
+  title: "Hedera Agent Chat",
+  description: "Chat with a Hedera agent scaffolded by create-hedera-agent.",
+};
+
+export default function RootLayout({ children }) {
+  // Read the operator key server-side and pass it to the browser-side wallet
+  // simulator. The key is intentionally serialized into the initial HTML.
+  // That's the "simulation" part. A real wallet holds its own key and the
+  // server never sees it; swapping <ChatWalletProvider> for a wallet SDK is
+  // the upgrade path.
```
skurzyp-blockydevs, line 27: so we now do not support the BrowserWallet extensions - like the Hashpack. I belive the previous nextJS example supported logging with hashpack.

packages/create-hedera-agent/template/web/src/features/chat-hedera/components/ChatHederaModeToggle.jsx
```diff
@@ -0,0 +1,94 @@
...
+  auto: "Auto",
+};
+
+// Title + body shown in the hover popover. Title repeats the button label so
+// the popover reads cleanly without the user having to map it back to the
+// button it belongs to.
+const MODE_DESCRIPTION = {
+  human: {
+    title: "Human approval",
+    body: "Mutating transactions pause and open the wallet drawer. You sign each one in-browser before it's broadcast. Nothing executes without your explicit OK.",
```
skurzyp-blockydevs: Consider adding info that queries will be executed directly as it might not be clear. In fact, all transactions will be pausing the execution loop - we have only mutating txs or queries.

packages/create-hedera-agent/template/web/src/features/chat-hedera/server/mutating-tools.js
```diff
@@ -0,0 +1,24 @@
...
+  const methods = new Set();
+  for (const plugin of plugins) {
+    if (isQueryPlugin(plugin)) continue;
+    const tools = plugin.tools({});
+    for (const tool of tools) {
+      methods.add(tool.method);
+    }
+  }
+  return methods;
+}
```
skurzyp-blockydevs, line 16: What about third party plugins? They might not follow this convention of naming

packages/create-hedera-agent/template/web/src/features/chat-hedera/server/submit-signed.js
```diff
@@ -0,0 +1,76 @@
...
+        accountId: receipt.accountId?.toString(),
+        tokenId: receipt.tokenId?.toString(),
+        topicId: receipt.topicId?.toString(),
+        scheduleId: receipt.scheduleId?.toString(),
+      },
+      humanMessage: `Transaction submitted: ${receipt.status.toString()}.`,
+    };
+    return Response.json(envelope, { status: 200 });
+  } catch (error) {
+    const message = error instanceof Error ? error.message : String(error);
```
skurzyp-blockydevs, line 55: This might not work correctly for third party plugins.
Consider the following tool:
https://github.com/jmgomezl/hak-saucerswap-plugin/blob/main/src/tools/swap.ts
1. the naming of the plugin wont be detected as mutating tool
2. even if it was, the tool originally (in AUTONOMUS mode) will trigger `handleTransaction` and then extend the response with additional information. 
```ts
  async secondaryAction(payload: SwapCorePayload, client: Client, context: Context) {
    const result = await handleTransaction(
      payload.transaction,
      client,
      context,
      swapPostProcess,
    );
    return { ...result, ...payload.extras };
  }
```

This appears to me like a potential problem of the return bytes mode and it integrations. We might need to rethink how it can be better handled. In case of this app this cannot be currently resolved but should be described in the readme and/or some UI toast/modal with info that third party plugins responses migth be poor.

packages/create-hedera-agent/template/web/src/features/chat-hedera/server/toolkit.js
```diff
@@ -0,0 +1,179 @@
...
+        const payload = {
+          raw: {
+            status: AWAITING_APPROVAL_STATUS,
+            toolName: name,
+            input,
+            summary: summarize(name, input),
+            unsignedBytes,
+          },
+          humanMessage:
+            "Awaiting user approval. The user must sign the transaction externally and submit the signed bytes. Do not call any further tools.",
```
skurzyp-blockydevs, line 95: I think this is risky. I dont understand fully the approach for the ReAct loop here but if we assume that after this is returned to the agent, the agent will obey and wait thats risky.

Also, the HITL mode should just pause the ReAct loop when bytes exist in the response of a tool call. Maybe we dont need to wrap the tools at all? We just check if tool returned bytes and if so then we intercept the execution and prompt user for signing.
The wrapping approach seems very complex to me

packages/create-hedera-agent/template/web/src/features/chat-hedera/utils/agent-mode.js
skurzyp-blockydevs, line 1: is this file necessary?

packages/create-hedera-agent/template/web/src/features/chat/state/chat-store.js
```diff
@@ -0,0 +1,299 @@
...
+  if (!raw) return [];
+  const parsed = safeParse(raw);
+  if (!Array.isArray(parsed)) return [];
+  const entries = parsed
+    .map(normalizeIndexEntry)
+    .filter((e) => e !== null);
+  return sortIndexEntries(entries);
+}
+
+export function loadChat(adapter, id) {
```
skurzyp-blockydevs, line 30: this will load a chat history for the user but will the Agent be aware of it?

skurzyp-blockydevs (changes requested):

packages/create-hedera-agent/template/web/src/app/layout.jsx
kierzniak: I don't want to make this example more complicated. I was inspired with AgentLab chatbot where there is not wallet integration and we only click button to sign transaction.

packages/create-hedera-agent/template/web/src/features/chat-hedera/server/mutating-tools.js
kierzniak: For now we do not support third party plugins and it is mentioned in readme.

packages/create-hedera-agent/template/web/src/features/chat-hedera/server/submit-signed.js
kierzniak: For now we do not support third party plugins and it is mentioned in readme.

packages/create-hedera-agent/template/web/src/features/chat/state/chat-store.js
kierzniak: Yes chat is aware of history.

packages/create-hedera-agent/template/shared/config.js
skurzyp-blockydevs: I see tht readme mentiones only ECDSA in both hex and DER encoding

packages/create-hedera-agent/template/web/src/app/layout.jsx
skurzyp-blockydevs: Okey, I get it. The root cause of us not integrating the wallet into the AgentLab was due to the fact that Kabila team mentioned working on sth similar. I think that using the wallet connect where it is possible - like here - should be a default solution over simulating the wallet.
The up side of your approach is that it does not require the WALLET_CONNECT_PROJECT_ID env file so technically one step less for user to configure it.
I think we should ask @piotrswierzy about his opinion.

packages/create-hedera-agent/template/shared/config.js
kierzniak: I introduced some changes and tested:
* Hex ECDSA key
* DER ECDSA key
* DER ED25519 key

All worked.

packages/create-hedera-agent/template/web/src/features/chat-hedera/components/ChatHederaModeToggle.jsx
kierzniak: Added

packages/create-hedera-agent/template/web/src/features/chat-hedera/utils/agent-mode.js
kierzniak: Deleted

packages/create-hedera-agent/template/web/src/app/layout.jsx
kierzniak: I think we can leave it like this for now and add support for real wallets features in the future

packages/create-hedera-agent/template/web/src/features/chat-hedera/server/toolkit.js
kierzniak: Prompt is only informative, HITL feature is based on `stopWhen` options in `ai-sdk`. Basically when we call any mutation tool in human mode agent will stop executing:

https://github.com/hashgraph/hedera-agent-kit-js/pull/819/changes#diff-47359d5e89b58b81848302665cd6a247eb408883e2a3febe8c9415ab2f058580R45

skurzyp-blockydevs (dismissed):

PR #845, opened 2026-05-23
docs: document return-bytes safety default

### LINKED ISSUE #841: Make return-bytes and human approval examples more prominent in the quick start
# Hedera Agent Kit Feedback Draft

Post this as a GitHub issue or short X thread before submitting the bounty form.

Title: Make return-bytes and human approval examples more prominent in the quick start

Feedback:

While building a small Week 1 Hedera AI Bounty agent, the Agent Kit quick start made it easy to wire a Hedera SDK client, `allCorePlugins`, and `HederaLangchainToolkit` into a LangChain agent. The main place I had to slow down was safety. The docs mention `AgentMode.RETURN_BYTE`, but the quick-start example uses autonomous mode, and a new builder might copy that into a fund-moving demo before thinking through consent and mainnet risk.

Suggested improvement:

- Add a "safe default" quick-start variant that uses `AgentMode.RETURN_BYTE`.
- Show an explicit approval step for transfers and other paid ledger writes.
- Include a small prompt classifier or policy snippet that blocks transfer-like requests until the operator confirms amount, recipient, memo, network, and fee.
- Call out the recommended default for bounty demos and public examples.

This would make it easier for first-time Agent Kit users to build impressive demos without accidentally creating agents that can move user funds without review.

pntech20:
## Summary
- update the quick-start examples to use `AgentMode.RETURN_BYTES` with an explicit account id
- add a small approval guard pattern for transfer-like prompts before agent invocation
- clarify the execution-mode docs so return-bytes mode is described as a review/signing flow, not automatic submission

## Checks
- git diff --check

Closes #841

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,7 +174,10 @@ async function main() {
     client,
     configuration: {
       plugins: allCorePlugins,
-      context: { mode: AgentMode.AUTONOMOUS },
+      context: {
+        mode: AgentMode.RETURN_BYTES,
+        accountId: process.env.ACCOUNT_ID!,
+      },
     },
   });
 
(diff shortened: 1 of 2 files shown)


skurzyp-blockydevs:
Hey @pntech20, thanks for your contribution! We will take a look at it and decide whether this is a direction we want to take. We will get back to you with our feedback and suggestions.

Please note that it may take some time as we have limited resources.

pntech20:
Title and DCO are fixed now. The remaining failing gate is the assignee check; GitHub does not allow contributors to self-assign this PR, so a maintainer will need to add the assignee.

skurzyp-blockydevs:
@pntech20 thank you for your contribution. I want to express my doubts about this approach:
1. the risk mentioned by @jakub8 in https://github.com/hashgraph/hedera-agent-kit-js/issues/841 is reffering to unintended. mainnet spendings when following the readme. Note that the readme explicitly documents creation of an agent for testnet only: 
```ts
  // Hedera client setup (Testnet by default)
  const client = Client.forTestnet().setOperator(
    process.env.ACCOUNT_ID!,
    PrivateKey.fromStringECDSA(process.env.PRIVATE_KEY!)
  );
  ```
  which makes it impossible to spend anything on the mainnet.
  2. I would argue that spending tokens on testnet is risky. This is exaclty the place to experiment and we nowhere encourage to configuring the agent on the mainnet.
  3. The idea of the 60s example from readme is to show the fully agentic example that will quickly showcase the capabilities of the agent and the HAK SDK. Autonomus mode is the default mode of the hedera agent kit. Return byte modes has multiple examples which are pointed by the docs: https://github.com/hashgraph/hedera-agent-kit-js/blob/main/docs/DEVEXAMPLES.md
  
  Thank you for your contributions and we will be happy to hear what do you think about our approach to this matter. I'm suggesting to leave the readme as it is now.

PR #863, opened 2026-05-25
feat: add read-only toolkit option

### LINKED ISSUE #861: Add a `readOnly: true` toolkit option to filter out write tools at construction time
### Contact Info

Hebx

### Describe your feedback, bug, or feature request

 While building HashTrail (a Hedera receipt agent that exposes both a deterministic command CLI and a free-form Q&A path), I wanted to give an LLM-backed ReAct agent access to live Hedera state without ever giving it a tool that could spend HBAR, mint NFTs, or submit HCS messages.

 Today the cleanest way to do that is to instantiate a `HederaLangchainToolkit`, then post-filter `toolkit.getTools()` for the `get_*` prefix before passing them to `createAgent`. That works, but it has two downsides:

> 1. The filter happens after the kit constructs every tool, so write-capable client wiring is built and discarded for nothing.
> 2. New write tools added in future plugins are silently inherited unless the consumer remembers to update their allowlist regex.
>
 A `readOnly: true` option on the toolkit (or a curated `coreReadOnlyPlugin` bundle) would let consumers express the intent declaratively. Something like:

> ```ts
> const toolkit = new HederaLangchainToolkit({
>   client,
>   configuration: {
>     plugins: [coreAccountQueryPlugin, coreConsensusQueryPlugin, coreTokenQueryPlugin, ...],
>     readOnly: true,
>     context: { mode: AgentMode.AUTONOMOUS },
>   },
> });
> ```

The toolkit could then assert each registered plugin only contributes `get_*` / non-mutating tools, and throw at construction time if a write plugin is mixed in. That gives Agent Kit users a hard, type-system-visible guarantee for the read-only Q&A pattern, which is exactly the safety surface the bounty rules ask for ("impossible to use or drain a user's funds without their explicit consent").

> Concrete examples in HashTrail:
>
> - `src/agent/free-form-agent.ts` — current post-filter approach, would simplify to `readOnly: true`.
> - `src/hedera/agent-kit.ts` — write toolkit stays as-is for the deterministic command path.
>
> Repo: https://github.com/Hebx/hashtrail-hedera-agent (v0.2.0).

Happy to send a PR if the maintainers agree on the shape.

### What type of feedback is this?

- [x] General Feedback
- [ ] Bug or Suggestion
- [x] Feature Request
- [ ] Other

### What does this feedback apply to?

- [x] Hedera Agent Kit JS
- [ ] Hedera Agent Kit Python
- [x] Feature Request
- [ ] Other

### What you expected vs what you experienced

**Expected:** When building a free-form Q&A agent on top of `HederaLangchainToolkit`, I expected a first-class way to say "this agent must be read-only" — either a `readOnly: true` flag on the toolkit configuration, or a curated `coreReadOnlyPlugin` bundle that mirrors the existing core plugins but contains only `get_*` tools. The intent is the same one the bounty rules call out: "impossible to use or drain a user's funds without their explicit consent." That's a common enough safety surface that I expected the kit to express it declaratively.

**Experienced:** I had to assemble it manually:

1. Mix the query and write plugins on the toolkit (`coreAccountQueryPlugin`, `coreConsensusPlugin`, `coreConsensusQueryPlugin`, `coreTokenPlugin`, `coreTokenQueryPlugin`, `coreTransactionQueryPlugin`, `coreMiscQueriesPlugin`).
2. Call `toolkit.getTools()`.
3. Post-filter the array with a regex/allowlist (`tool.name.startsWith("get_")` plus a few explicit exceptions) before handing the result to `createAgent`.

This works, and it's what HashTrail v0.2.0 ships in [`src/agent/free-form-agent.ts`](https://github.com/Hebx/hashtrail-hedera-agent/blob/main/src/agent/free-form-agent.ts), but it has three real costs:

- **Wasted construction.** Every write-capable tool is fully built (operator client wiring, schema, signer hooks) and immediately discarded.
- **Silent regression risk.** If a future plugin version adds a new write tool whose name doesn't match the consumer's allowlist regex, it gets handed to the LLM unnoticed. The safety contract leaks out of the toolkit and into every consumer's filter.
- **No type-system guarantee.** The "this agent cannot spend" property lives in a runtime regex, not in the toolkit type. There's no way for the compiler to fail a build that mixes a write plugin into a "read-only" agent.

A `readOnly: true` option (or `coreReadOnlyPlugin`) that asserts at construction time and either rejects or filters write-capable tools would push that contract into the kit, where it belongs.

### Relevant links or resources

- HashTrail repo (v0.2.0): https://github.com/Hebx/hashtrail-hedera-agent
- Current read-only filter implementation: https://github.com/Hebx/hashtrail-hedera-agent/blob/main/src/agent/free-form-agent.ts
- Write-side toolkit boundary (kept separate, with mint allowlist + 1 HBAR cap): https://github.com/Hebx/hashtrail-hedera-agent/blob/main/src/hedera/agent-kit.ts
- Architecture diagram showing the dual routing (yellow = gated write paths, blue = read-only Q&A path): https://github.com/Hebx/hashtrail-hedera-agent#architecture
- v0.2.0 release notes: https://github.com/Hebx/hashtrail-hedera-agent/releases/tag/v0.2.0
- Hedera AI Bounty week 1 rule this addresses: "Mainnet agents must use a human-in-the-loop flow, or equivalent safeguards against misuse of funds."

### Score (1-10): How confident did you feel after reading the docs that you could build successfully?

7

### Score (1-10): How easy was it to get help when you were blocked?

8

### Score (1-10): How intuitive were the APIs / SDKs to use?

9

### Score (1-10): How easy was it to debug issues?

9

### Score (1-10): How likely are you to build again on Hedera again?

10

Lucas-FManager:
## Summary
- Add a `readOnly` option to core configuration.
- Enforce read-only tool discovery by rejecting non-query tools when `readOnly` is enabled.
- Add unit coverage for read-only, mixed-tool rejection, and existing default behavior.

## Approach
The guard is intentionally fail-closed: if a plugin returns any non-`get_` tool while `readOnly` is enabled, tool discovery throws with the unsafe method names instead of silently exposing write-capable actions to an agent.

## Validation
- `pnpm exec vitest run tests/unit/shared/tool-discovery.unit.test.ts`

Note: the package `test:unit` script also ran the broader unit suite locally and surfaced an existing package-entry resolution issue in `tests/unit/utils/decimals-utils.unit.test.ts`; the new `tool-discovery` unit test passed in that run and passes directly with the command above.

Closes #861

Disclosure: this PR was prepared with AI assistance through Codex. I reviewed the diff and validation before submitting it under my GitHub account.

### DIFF
--- a/packages/core/src/shared/configuration.ts
+++ b/packages/core/src/shared/configuration.ts
@@ -31,4 +31,6 @@ export type Configuration = {
   //plugins to load - must be explicitly provided
   plugins?: Plugin[];
   context?: Context;
+  // When true, only non-mutating query tools may be registered.
+  readOnly?: boolean;
 };
--- a/packages/core/src/shared/tool-discovery.ts
+++ b/packages/core/src/shared/tool-discovery.ts
@@ -12,6 +12,9 @@ export class ToolDiscovery {
   getAllTools(context: Context, configuration?: Configuration): Tool[] {
     // Get plugin tools
     const pluginTools = this.pluginRegistry.getTools(context);
+    if (configuration?.readOnly) {
+      this.assertReadOnlyTools(pluginTools);
+    }
 
     // Merge all tools (core tools take precedence in case of name conflicts)
     const allTools: any = [];
@@ -35,6 +38,20 @@ export class ToolDiscovery {
     }
 
     return allTools;
+  }
+
+  private assertReadOnlyTools(tools: Tool[]): void {
+    const writeTools = tools.filter(tool => !this.isReadOnlyTool(tool));
+    if (writeTools.length === 0) {
+      return;
+    }
+
+    const methods = writeTools.map(tool => tool.method).join(', ');
+    throw new Error(`readOnly configuration rejects write-capable tools: ${methods}`);
+  }
+
+  private isReadOnlyTool(tool: Tool): boolean {
+    return tool.method.startsWith('get_');
   }
 
   static createFromConfiguration(configuration: Configuration): ToolDiscovery {
--- a/packages/core/tests/unit/shared/tool-discovery.unit.test.ts
+++ b/packages/core/tests/unit/shared/tool-discovery.unit.test.ts
@@ -1,0 +1,57 @@
+import { describe, expect, it } from 'vitest';
+import { z } from 'zod';
+import { ToolDiscovery } from '@/shared/tool-discovery';
+import type { Context } from '@/shared/configuration';
+import type { Plugin } from '@/shared/plugin';
+import type { Tool } from '@/shared/tools';
+
+const makeTool = (method: string): Tool => ({
+  method,
+  name: method,
+  description: `${method} description`,
+  parameters: z.object({}),
+  execute: async () => ({}),
+});
+
+const makePlugin = (name: string, methods: string[]): Plugin => ({
+  name,
+  tools: (_context: Context) => methods.map(makeTool),
+});
+
+describe('ToolDiscovery readOnly configuration', () => {
+  it('keeps query tools when readOnly is enabled', () => {
+    const discovery = new ToolDiscovery([
+      makePlugin('queries', ['get_account_query_tool', 'get_hbar_balance_query_tool']),
+    ]);
+
+    const tools = discovery.getAllTools({}, {readOnly: true});
+
+    expect(tools.map(tool => tool.method)).toEqual([
+      'get_account_query_tool',
+      'get_hbar_balance_query_tool',
+    ]);
+  });
+
+  it('rejects write-capable tools when readOnly is enabled', () => {
+    const discovery = new ToolDiscovery([
+      makePlugin('mixed', ['get_account_query_tool', 'transfer_hbar_tool']),
+    ]);
+
+    expect(() => discovery.getAllTools({}, {readOnly: true})).toThrow(
+      /transfer_hbar_tool/,
+    );
+  });
+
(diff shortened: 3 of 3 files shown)


Lucas-FManager:
The code checks are green except for the required Assignee Check, which fails with \Assignee is not set\. I attempted to assign myself, but GitHub rejected it because I do not have assign permissions on this repository. Could a maintainer please assign this PR or the linked issue when appropriate?

skurzyp-blockydevs:
Hi @Lucas-FManager! Thank you for your contribution. We will review it and get back to you. Please note that it may take some time due to limited resources.

Lucas-FManager:
Thanks for the update. I will keep the PR ready and wait for maintainer review.

skurzyp-blockydevs:
Hey @Lucas-FManager, we decided to go a different way. Your solution duplicates information that is already available. Devs can easily filter out the given type of tools themself or just import the query plugins. Our take is that the approach based on the names is working but is not very reliable. Thats we decided to add a tool type field to the tool classes - this will explicitly define if tool is a `transaction` or `query` tool - i other cases and for third party tools the type will be `other`. This would simplify the filtering and could be useful not only here but also in the hooks and policies. 

I'm closing this PR as we decided to not pursue this approach. Thank you for the suggestion and bringing that problem to our attention.

There is a new issue covering our aproach idea available here: https://github.com/hashgraph/hedera-agent-kit-js/issues/1002

PR #893, opened 2026-06-03
feat(core): add minimal DEX swap_tokens_tool (#7)

jmgomezl:
## What this does

Adds a minimal token swap tool, addressing #7.

It introduces a small `core-dex-plugin` with a single `swap_tokens_tool` that calls a Uniswap V2 style router's `swapExactTokensForTokens(amountIn, amountOutMin, path, to, deadline)`. It's deliberately vendor-neutral: you pass the router contract, the token route, the amounts, and an optional deadline, so it works with any compatible Hedera DEX (SaucerSwap and similar) without baking in a specific provider.

I kept the surface area intentionally small, in line with the "minimal proof of concept" note on the issue.

## How it works

- New schema `swapExactTokensForTokensParameters` and a `normaliseSwapExactTokensForTokensParams` normaliser that resolves the route + recipient to EVM addresses, encodes the call with `ethers`, and defaults the deadline to 20 minutes out.
- The tool follows the same shape as the existing ERC-20 tools (ECDSA operator assertion, `HederaBuilder.executeTransaction`, mode-aware `handleTransaction`).
- Registered in the plugins barrel and documented in `HEDERATOOLS.md` / `HEDERAPLUGINS.md`.

## A note on allowances

The input token needs an allowance granted to the router before swapping. Rather than bundle that in, the tool documents the prerequisite and points to the existing `approve_token_allowance_tool`, keeping each tool focused.

## Testing

- `pnpm lint:check`, `pnpm build`, and `pnpm test:unit` all pass.
- New unit test covers metadata, the happy path, and error handling.

## Possible follow-ups (out of scope here)

- A dedicated ERC-20 `approve` tool (core currently only has the HTS allowance tool).
- Quote/slippage helpers and multi-hop routing convenience.

Happy to adjust the naming, parameters, or structure to fit your conventions. Thanks for taking a look!

### DIFF
--- a/docs/HEDERAPLUGINS.md
+++ b/docs/HEDERAPLUGINS.md
@@ -18,6 +18,7 @@ The tools are organized into plugins, each containing related functionality:
 * **Core Token Plugin**: Tools for Hedera Token Service (HTS) operations
 * **Core Token Query Plugin**: Tools for querying Hedera Token Service related data
 * **Core EVM Plugin**: Tools for interacting with EVM smart contracts on Hedera (ERC-20 and ERC-721)
+* **Core DEX Plugin**: Tools for swapping tokens on Hedera DEXs (Uniswap V2 style routers)
 * **Core EVM Query Plugin**: Tools for querying smart contract-related data on Hedera
 * **Core Misc Query Plugin**: Tools for fetching miscellaneous information from Hedera Mirror Node
 * **Core Transaction Query Plugin**: Tools for handling Hedera transaction–related queries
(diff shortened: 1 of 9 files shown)


jmgomezl:
Pushed a validation hardening commit while reviewing the Wiz SAST finding on this PR.

## What was wrong

`amountIn` and `amountOutMin` were plain `z.string()` and fed straight into `BigInt()` during normalisation. I tested what that actually accepted:

| Input | Schema | `BigInt()` |
|---|---|---|
| `"100"` | accept | `100` ✅ |
| `"0"` | accept | `0` ⚠️ |
| `""` | accept | **`0`** ⚠️ |
| `"-5"` | accept | `-5` ⚠️ |
| `"abc"` / `"1.5"` / `"1e10"` | accept | **throws `SyntaxError`** |

The one that matters: **`amountOutMin` accepted `"0"` and `""`.** That is the slippage floor — a zero floor accepts any output amount, leaving the swap open to sandwiching. Since this is an LLM-driven tool, a fumbled unit conversion could silently disable the only economic protection in the call, and the `""` case failed silently rather than visibly.

`"1.5"` and `"1e10"` are also exactly what a model emits when it ignores "base units, integer string", and they surfaced as a raw `SyntaxError` from inside the normaliser rather than a usable validation error.

## The fix

Both fields are now constrained to positive integer strings at the schema boundary.

One implementation note worth flagging: I first wrote this as `.regex(...).refine(v => BigInt(v) > 0n)`, and my own new tests caught that **zod still runs a refinement after an earlier regex check has failed** — so `BigInt("1e10")` threw inside the refinement, exactly on the inputs the regex existed to reject. It is now two plain string regexes, so no code path can throw.

Adds 11 unit tests: zero, empty, negative, non-integer, and above-`MAX_SAFE_INTEGER` amounts.

## Verification

Full core unit suite: **510 passed** (up from 499, +11 new), no regressions. ESLint clean. I also ran prettier on both files — they were unformatted when originally added, and both are new in this PR, so it touches nothing outside my own changes.

The 1 pre-existing test-file collection error (a Vite `normalizeUrl` internal) is unrelated — I confirmed it on a clean checkout too.

## Still needs a maintainer

I could not read the actual Wiz SAST finding — it posted 0 GitHub annotations, so the detail is only in the `app.wiz.io` dashboard. **The above is my own review, not necessarily what Wiz flagged.** If someone can paste the finding I will address the real one. For reference it is reported as `NEUTRAL` / 1 Medium, so it should not be blocking.

`Assignee Check` is still failing here too — outside contributors cannot self-assign.

Issue #837, opened 2026-05-21
DX feedback: @langchain/core version pin, HcsAuditTrailHook ergonomics, and streaming/tool-name gaps


hcharper:
Feedback after building a full agent (HBAR / HTS / HCS) on the JS Agent Kit v4. Loved how fast `allCorePlugins` + the LangChain toolkit got me to real transactions on testnet. A few friction points worth sharing:

1. **`@langchain/core` version conflict.** `@hashgraph/hedera-agent-kit-langchain` pins `@langchain/core@1.1.24` exactly, while `langchain@1.x` wants `^1.1.47`. Two copies end up installed, and `toolkit.getTools()` no longer type-matches `createAgent`'s `tools` parameter — I had to cast through `unknown as ClientTool[]`. Loosening the pin to a range (e.g. `^1.1.24`) would avoid this.

2. **`HcsAuditTrailHook` requires a pre-existing topic.** It doesn't create one, and it throws outside `AUTONOMOUS` mode. A `createTopicIfMissing` option, or a small documented helper to bootstrap the topic, would smooth the setup.

3. **Streaming isn't documented.** The examples use `agent.invoke()`. For a live UI I needed `agent.streamEvents({ version: 'v2' })` — a streaming example plus the event shape (tool start/end, token deltas) would help a lot.

4. **Tool names in stream events are the generic class name.** `on_tool_start` reports `HederaAgentKitTool` rather than e.g. `transfer_hbar`; I had to extract the real tool name from the serialized `ToolMessage` on `on_tool_end` to label UI steps.

5. **Feature gaps for "do anything on Hedera" agents.** No smart-contract deploy/call, file service, or schedule-transaction tools in core — would be great additions.

Overall a great kit — these are all minor and the captain's-log audit hook in particular was a highlight. Thanks!

skurzyp-blockydevs:
@hcharper Hi! Really appreciate your detailed feedback. We will look into the pain points you mentioned and decide how to address them. Glad you like the SDK! Good luck with your project, and we’ll get back to you later. Please note that it may take some time as we have limited resources.

MWBlocky:
1. @langchain/core version conflict / duplicate copies - Fixed. 
The host-framework packages (@langchain/core, langchain, and the AI SDK / ADK / ElizaOS equivalents) are now peerDependencies instead of pinned dependencies, so they dedupe against your app's copy. No more double install or `as unknown as ClientTool[]` cast.

[...]

PR #974, opened 2026-07-01
docs(hooks): clarify HcsAuditTrailHook is AUTONOMOUS-only and needs a…

MWBlocky:
Related to: #837

## Summary

Adds JSDoc to `HcsAuditTrailHook` (`packages/core/src/hooks/hcs-audit-trail-hook.ts`) making its two setup requirements visible where developers actually meet them — the class and constructor, via IDE hover/autocomplete. Documentation only; no behavior change.

## Why

Feedback from a developer who built a full agent on the kit:

> `HcsAuditTrailHook` requires a pre-existing topic. It doesn't create one, and it throws outside `AUTONOMOUS` mode. A `createTopicIfMissing` option, or a small documented helper to bootstrap the topic, would smooth the setup.

Both constraints are real:

- The constructor takes a required `hcsTopicId` and never creates a topic.
- `preToolExecutionHook` throws when the agent runs in `RETURN_BYTES` mode — the hook is `AUTONOMOUS`-only.

The dedicated guide `docs/HOOKS_AND_POLICIES.md` **already** documents both (an "Autonomous Mode Only" note and a "Topic Creation: must be created before initializing the hook" prerequisite). The gap was the **source JSDoc**: it only warned about paid topics, so a developer using the hook through editor tooling — as the reporter did — never saw either constraint. This PR closes that gap rather than duplicating the prose docs.

## Changes

`packages/core/src/hooks/hcs-audit-trail-hook.ts`:

- Class JSDoc: added an `@remarks` block stating the hook is **AUTONOMOUS-only** (throws in `RETURN_BYTES`) and **requires a pre-existing topic** (create it via `create_topic_tool` / `TopicCreateTransaction` and pass its id; the client must be allowed to submit). Kept the existing HIP-991 paid-topic warning and added an `@see` link to `docs/HOOKS_AND_POLICIES.md`.
- Constructor JSDoc: `@param` descriptions for `relevantTools`, `hcsTopicId` (noting it must be **pre-existing**), and `loggingClient`.

## Scope / follow-ups

This is the documentation-only remediation for the feedback. The reporter also floated a `createTopicIfMissing` option (auto-creating the topic) — deliberately **out of scope** here; it's a behavior change that can be considered separately if there's demand.

## Testing

None required — comment-only change, no runtime or type impact.

### DIFF
--- a/packages/core/src/hooks/hcs-audit-trail-hook.ts
+++ b/packages/core/src/hooks/hcs-audit-trail-hook.ts
@@ -10,8 +10,18 @@ import { Client, TopicMessageSubmitTransaction } from '@hiero-ledger/sdk';
 /**
  * Hook to add an audit trail of tool executions to a Hedera Consensus Service (HCS) topic.
  *
+ * @remarks
+ * **Autonomous mode only.** This hook works only when the agent runs in `AgentMode.AUTONOMOUS`.
+ * In `RETURN_BYTES` mode it throws before the tool executes, because there is no submitted
+ * transaction to audit.
+ *
+ * **Requires a pre-existing topic.** The hook does not create an HCS topic. Create the topic
+ * first — via the `create_topic_tool`.
+ *
  * @warning If a paid topic (HIP-991: https://hips.hedera.com/hip/hip-991) is provided,
  * it could potentially drain the provided logging client's funds due to message submission fees.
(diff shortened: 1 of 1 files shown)


skurzyp-blockydevs (approved):
LGTM!

PR #977, opened 2026-07-02
feat(langchain): expose tool name in stream events via event.metadata.hakToolName

MWBlocky:
**Description**:
`on_tool_start.event.name` reports the wrapper class `HederaAgentKitTool` 
this adds the real name at step start as `event.metadata.hakToolName`. 

**Related issue(s)**:

Related to #837

### DIFF
--- a/packages/langchain/README.md
+++ b/packages/langchain/README.md
@@ -78,6 +78,18 @@ console.log(parsed[0].toolName);   // e.g. "create_account_tool"
 console.log(parsed[0].parsedData); // tool execution result
 ```
 
+## Streaming
+
+When streaming with `agent.streamEvents({ version: 'v2' })`, the `on_tool_start` event's `name` is the shared wrapper class (`HederaAgentKitTool`). To label steps with the real tool name, read it from `event.metadata.hakToolName`:
+
+```typescript
+for await (const event of agent.streamEvents({ messages }, { version: 'v2' })) {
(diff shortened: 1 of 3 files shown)


packages/langchain/README.md
```diff
@@ -78,6 +78,18 @@ console.log(parsed[0].toolName);   // e.g. "create_account_tool"
...
+When streaming with `agent.streamEvents({ version: 'v2' })`, the `on_tool_start` event's `name` is the shared wrapper class (`HederaAgentKitTool`). To label steps with the real tool name, read it from `event.metadata.hakToolName`:
+
+```typescript
+for await (const event of agent.streamEvents({ messages }, { version: 'v2' })) {
+  if (event.event === 'on_tool_start') {
+    console.log(event.metadata?.hakToolName); // e.g. "transfer_hbar_tool"
+  }
+}
+```
+
```
skurzyp-blockydevs, line 92: I think that we should add an example langchain/vercelAI agent that supports streaming response and document that in `@docs/DEVEXAMPLES.md`. This can be done as a separate PR tho

skurzyp-blockydevs (approved):

packages/langchain/README.md
MWBlocky: -> https://github.com/hashgraph/hedera-agent-kit-js/issues/981

PR #978, opened 2026-07-02
fix(mirrornode): retry transaction-record 404s with backoff

### LINKED ISSUE #857: Feedback from building Kalipso (Week 1 bounty): v4 SDK transition friction, mirror-node tx-id gap, transitive vulnerabilities

MWBlocky:
**Description**:
A freshly submitted transaction is not indexed by the mirror node for a few seconds and returns 404, so the transfer -> verify-record handoff failed. Retry 404s up to 3 attempts with exponential backoff (1s, 2s). Non-404 errors and an exhausted 404 throw as before.

**Related issue(s)**:

Fixes #857


skurzyp-blockydevs (changes requested):
if we want to implement retry with a backoff approach we should incorporate it consistently to each of the mirrornode methods. The same thing might happen if get token info is fired too early after the token creation. This means we probably should have some middleware for each fetch done by the mirrornode service. What do you think about this idea?

skurzyp-blockydevs (approved):
lgtm

PR #983, opened 2026-07-03
feat: typed tool result status

### LINKED ISSUE #987: feat: standarise tool status reporting and implement classifyToolResult helper
This issue is based on the feedback in: https://github.com/hashgraph/hedera-agent-kit-js/issues/967 and PR: https://github.com/hashgraph/hedera-agent-kit-js/pull/968

The scope of changes extends introduction of the `classifyToolResult` helper for typed tool result envelopes that was added in #968.

skurzyp-blockydevs:
# Tool Result Status

### New: `TOOL_STATUS` const — single source of truth for status strings

A shared constant replaces every bare `'SUCCESS'` / `'ERROR'` / `'PARSE_ERROR'` string literal at every write and read site:

```typescript
export const TOOL_STATUS = {
  SUCCESS: 'SUCCESS',
  ERROR: 'ERROR',
  PARSE_ERROR: 'PARSE_ERROR',
} as const;

export type ToolRawStatus = (typeof TOOL_STATUS)[keyof typeof TOOL_STATUS];
```

`ToolRawStatus` is now derived from the const, so the type and the runtime values are always in sync. A typo like `TOOL_STATUS.SUCESS` is a compile error; the string `'SUCESS'` was not. `TOOL_STATUS` is exported from `@hashgraph/hedera-agent-kit` for plugin authors.

---

### New: `classifyToolResult` and `ToolResultStatus<T>` — created by @sneldao

`classifyToolResult` takes the `{ raw, humanMessage }` envelope returned by `transactionToolOutputParser` or `untypedQueryOutputParser` and maps it to a typed discriminated union:

```typescript
type ToolResultStatus<T = unknown> =
  | { kind: 'success';     transactionId?: string; data: T; humanMessage: string }
  | { kind: 'failure';     errorCode: string; transactionId?: string; error: string; humanMessage: string }
  | { kind: 'parse_error'; originalOutput: unknown; humanMessage: string }
  | { kind: 'unknown';     humanMessage: string }
```

Classification rules:

| `raw.status` | `raw.errorCode` | `raw.error` | `kind` | `errorCode` |
|---|---|---|---|---|
| `'SUCCESS'` | — | — | `success` | — |
| `'ERROR'` | `'INSUFFICIENT_PAYER_BALANCE'` (SDK status string) | string | `failure` | `'INSUFFICIENT_PAYER_BALANCE'` |
| `'ERROR'` | absent | string | `failure` | `'ERROR'` |
| `'PARSE_ERROR'` | — | — | `parse_error` | — |
| anything else | — | absent | `unknown` | — |
| `null` / non-object | — | — | `parse_error` | — |

`errorCode` is populated from `raw.errorCode` when present (the specific Hedera SDK status name for receipt/precheck failures), otherwise falls back to `raw.status`. `transactionId` is lifted out of `raw` for both `success` and `failure` kinds — receipt and precheck errors both set it.

`kind: 'unknown'` exists specifically to prevent unrecognized shapes from silently falling through to `success`.

---

### Changed: `status: 'SUCCESS'` is now defaulted centrally in `BaseTool.execute`

`classifyToolResult` requires `raw.status === 'SUCCESS'` to return `kind: 'success'`. Instead of requiring every tool to set it explicitly, `BaseTool.execute()` now defaults it with `??=` after the success pipeline completes:

```typescript
// In BaseTool.execute(), after secondaryAction:
if (result && typeof result.raw === 'object' && result.raw !== null) {
  result.raw.status ??= TOOL_STATUS.SUCCESS;
}
```

`??=` preserves any explicit status already set upstream (`ExecuteStrategy`, `ReturnBytesStrategy`). Query tools and third-party tools that return a `raw` envelope without a `status` field now get `'SUCCESS'` for free.

---

### Changed: `handleError` standardized to `status: 'ERROR'`

All `handleError` implementations previously used `Status.InvalidTransaction` (an SDK receipt-level object) for pre-flight errors — errors that occur before a transaction is ever submitted. This was semantically wrong and inconsistent after JSON round-trip (`Status.InvalidTransaction` serializes as `{ _code: 1 }`, not as a string).

All tools now return a plain string via `TOOL_STATUS.ERROR`:

```typescript
return { raw: { status: TOOL_STATUS.ERROR, error: message }, humanMessage: message };
```

Affected:
- `BaseTool.handleError` (base class)
- All 10 query tool `handleError` overrides — **deleted** (see below)
- All 33 transaction tool `handleError` overrides — **deleted** (see `BaseTransactionTool` below)

The unused `Status` import was removed from all affected files.

---

### New: `BaseTransactionTool` — structured `ReceiptStatusError` and `PrecheckStatusError` fields

A new abstract class `BaseTransactionTool extends BaseTool` was introduced in `packages/core/src/shared/base-transaction-tool.ts`. It owns the Hedera-specific error handling that previously lived (in an incomplete, prose-only form) in each of the 33 transaction tools.

**Why**

- `BaseTool` is a generic framework primitive that third-party plugins can extend. Baking Hedera SDK knowledge into it leaked SDK types into the generic base and would force every plugin author to depend on `@hiero-ledger/sdk` transitively.
- Before, when a transaction failed on-network or at the precheck stage, callers received a prose `humanMessage` and had to regex-parse it to recover the SDK status name and transaction ID. Both are already structured on the error — they should be surfaced directly.
- `INSUFFICIENT_PAYER_BALANCE` typically surfaces as a `PrecheckStatusError` (the node rejects the transaction before it reaches consensus), not a `ReceiptStatusError`. The previous implementation only handled `ReceiptStatusError`, so `errorCode` and `transactionId` were silently lost for the most common balance failure.
- The 33 transaction tools each carried a near-identical `handleError` override. That's boilerplate duplication and an easy source of drift.

**What the class does**

When `execute()` throws a `ReceiptStatusError` or `PrecheckStatusError`, `BaseTransactionTool.handleError`:

- Sets `raw.status = TOOL_STATUS.ERROR`
- Sets `raw.errorCode = error.status.toString()` — the SDK status **name**, e.g. `'INSUFFICIENT_PAYER_BALANCE'`, `'TOKEN_NOT_ASSOCIATED_TO_ACCOUNT'`, `'INVALID_SCHEDULE_ID'`
- Sets `raw.transactionId = error.transactionId.toString()` — e.g. `'0.0.9414979@1783085748.874255414'`
- Sets `raw.error = error.message` — the raw SDK message, no prefix, machine-readable
- Sets `humanMessage = "Failed to execute ${this.name}: ${error.message}"` — human-facing, prefixed with the tool name

For any other error type it falls through to `super.handleError(error, context)` (the generic `BaseTool` path).

**How it lands in the codebase**

- `BaseTool.handleError` no longer imports or references `ReceiptStatusError` or `PrecheckStatusError`. It's purely generic.
- All 33 transaction tools now `extends BaseTransactionTool` (instead of `BaseTool`) and their per-tool `handleError` overrides were removed — the shared implementation covers every case.
- Query tools and any other non-transaction tools continue to extend `BaseTool` directly. Their `handleError` overrides were also removed (they were identical to the base).
- Exports: `BaseTransactionTool` is re-exported from `packages/core/src/index.ts` for plugin authors.

**How to use the structured fields**

```typescript
const envelope = transactionToolOutputParser(rawOutput);
const result = classifyToolResult<{ transactionId: string }>(envelope);

switch (result.kind) {
  case 'success':
    return { txId: result.transactionId };
  case 'failure':
    // result.errorCode === 'INSUFFICIENT_PAYER_BALANCE' for receipt/precheck balance failures
    // result.transactionId is available for receipt/precheck failures
    // result.errorCode === 'ERROR' for generic pre-flight throws
    if (result.errorCode === 'INSUFFICIENT_PAYER_BALANCE') {
      await topUpPayer();
    }
    throw new Error(`tool failed (${result.errorCode}): ${result.error}`);
}
```

Or, without `classifyToolResult`, read `raw.errorCode` and `raw.transactionId` directly.

---

### Changed: query tool `handleError` overrides deleted

All 10 query tool `handleError` overrides were deleted — they were behaviorally identical to `BaseTool.handleError` after this PR's changes. Query tools now inherit the base implementation directly, mirroring what the transaction tools do via `BaseTransactionTool`.

Affected tools:
- `get-account-query`
- `get-account-token-balances-query`
- `get-hbar-balance-query`
- `get-token-info-query`
- `get-pending-airdrop-query`
- `get-exchange-rate-query`
- `get-transaction-record-query`
- `get-topic-info-query`
- `get-topic-messages-query`
- `get-contract-info-query`

---

### Changed: `raw.error` is the raw SDK message; `humanMessage` carries the prefix

Previously the `handleError` overrides on transaction tools set `raw.error` and `humanMessage` to the **same** prefixed string (e.g. `"Failed to transfer HBAR: receipt for transaction ... INSUFFICIENT_PAYER_BALANCE"`). After the `BaseTransactionTool` change, on the `ReceiptStatusError` / `PrecheckStatusError` path only:

- `raw.error` = the raw SDK message (`"receipt for transaction ... INSUFFICIENT_PAYER_BALANCE"`) — machine-readable, no tool-name prefix.
- `humanMessage` = the prefixed message (`"Failed to execute Transfer HBAR: receipt for transaction ... INSUFFICIENT_PAYER_BALANCE"`) — human-facing.

On the generic-error path (via `BaseTool.handleError`), both fields still carry the prefixed message.

---

### Changed: transaction tool `humanMessage` wording is now uniform

Because per-tool `handleError` overrides were removed, transaction failure `humanMessage`s now share a single template: `"Failed to execute ${this.name}: ${error.message}"`. Previously each tool phrased its own prefix (`"Failed to dissociate"`, `"Failed to transfer HBAR"`, `"Failed to sign scheduled transaction"`, etc.). The new form always starts with `"Failed to execute "` followed by the exact `name` field on the tool (e.g. `"Failed to execute Dissociate Token"`, `"Failed to execute Sign Scheduled Transaction"`).

Integration and e2e tests that asserted on the old per-tool prose were updated accordingly.

---

### Fixed: `convertBytes` in `responseParserService.ts` preserves sibling fields

The top-level-`bytes` branch of `convertBytes` previously returned only `{ bytes: … }`, discarding sibling fields including `status`. It now spreads `data` to preserve all fields:

```typescript
// Before
return { bytes: new Uint8Array(data.bytes.data) };
// After
return { ...data, bytes: new Uint8Array(data.bytes.data) };
```

This matches the behaviour of the nested `data.raw.bytes` branch, which already spread correctly.

---

### Fixed: migration doc example updated

The `handleError` example in `docs/MIGRATION-v4.md` previously showed `status: Status.InvalidTransaction`, which no longer matches the contract. It now shows `status: 'ERROR'`.

---

### External tools that do not set `status: 'SUCCESS'`

Any custom tool whose `coreAction` or `secondaryAction` returns a `raw` object without `status: 'SUCCESS'` will now have it automatically set to `'SUCCESS'` by `BaseTool.execute()` (via `??=`). This is the **desired** behaviour — the `??=` preserves any explicit status a tool does set. However, a tool that intentionally returns a non-SUCCESS non-ERROR status would have that status overwritten.
**Who is affected:** Plugin authors with custom `BaseTool` subclasses that intentionally return a non-standard `raw.status` value on the success path.
**Fix:** Use `TOOL_STATUS.ERROR` or `TOOL_STATUS.PARSE_ERROR` for non-success outcomes — the `??=` in `execute()` only fires when `raw.status` is `undefined` or `null`.

---

### External tools that override `handleError` with `Status.InvalidTransaction`

Third-party tools that override `handleError` and return `status: Status.InvalidTransaction` will produce `errorCode: 'UNKNOWN'` rather than `'ERROR'`. After a JSON round-trip `Status.InvalidTransaction` deserialises as `{ _code: 1 }` (an object, not a string), so `classifyToolResult` cannot match it against any known status string and falls back to `'UNKNOWN'`. The core tools no longer do this, but external plugins may still be inconsistent.
**Who is affected:** Plugin authors who copied the previous `handleError` pattern from core tools.
**Fix:** Replace `status: Status.InvalidTransaction` with `status: TOOL_STATUS.ERROR` in `handleError`.

---

### Consumers checking `raw.status._code` on error

Code that inspects `raw.status._code` to detect or branch on transaction tool errors will no longer find a numeric code — `raw.status` is now the string `'ERROR'`. The error description is in `raw.error` (and `humanMessage`).
**Who is affected:** Any consumer that was reading `raw.status._code` directly from a tool error envelope.
**Fix:** Use `classifyToolResult` and switch on `kind`. For `kind: 'failure'`, `errorCode` holds the value.

---

### `transactionToolOutputParser` RETURN_BYTES output shape

`raw` in RETURN_BYTES mode now includes `status: 'SUCCESS'` alongside `bytes` (set by `BaseTool.execute()` via `??=`, since `ReturnBytesStrategy` already sets it). Code that checked for the exact set of keys (e.g. `Object.keys(raw).length === 1`) would see an additional field.
**Who is affected:** Unlikely in practice, but any consumer parsing the RETURN_BYTES raw envelope by exact structure.
**Fix:** Check for the presence of `raw.bytes` rather than the absence of other keys.

---

### `raw.error` no longer carries the `"Failed to execute <Tool>: "` prefix on receipt/precheck failures

On a `ReceiptStatusError` or `PrecheckStatusError` path, `raw.error` is now the bare SDK message (e.g. `"receipt for transaction 0.0.9414979@... contained error status INSUFFICIENT_PAYER_BALANCE"`). Consumers previously seeing the prefixed form in `raw.error` and doing a `startsWith("Failed to ")` check will miss.
**Who is affected:** Anything that parses or asserts on the exact contents of `raw.error` for transaction tools.
**Fix:** If you need the human-facing prefixed string, read `humanMessage`. If you need to detect the SDK status, read `raw.errorCode` or `classifyToolResult(...).errorCode`. `raw.error` is now intended as the machine-readable underlying message.

Generic (non-receipt/precheck) errors are unchanged: `raw.error === humanMessage === "Failed to execute <Tool>: <message>"`.

---

### `humanMessage` wording changed for transaction failures

Every transaction tool's failure `humanMessage` now begins with `"Failed to execute <Tool Name>: ..."`. Previously each tool used a bespoke phrase (`"Failed to dissociate"`, `"Failed to sign scheduled transaction"`, `"Failed to update token:"`, etc.).
**Who is affected:** Any consumer, log filter, or test that regex-matches on the old per-tool prose.
**Fix:** Match on the new template. The `<Tool Name>` segment is exactly the tool's `name` field — see each tool file for the canonical string. Where possible, prefer branching on `raw.errorCode` or `classifyToolResult(...).errorCode` instead of `humanMessage`, since structured fields are stable across wording changes.

---

### Transaction tools now extend `BaseTransactionTool`, not `BaseTool`

All 33 in-repo transaction tools were reparented from `BaseTool` to `BaseTransactionTool`. The class hierarchy is `BaseTransactionTool extends BaseTool`, so `instanceof BaseTool` checks still succeed. However, custom transaction tools authored by plugin developers should also migrate to preserve the structured error fields.
**Who is affected:** Plugin authors with custom transaction tools that extend `BaseTool` directly and want the `errorCode` / `transactionId` promotion behavior.
**Fix:** Change `extends BaseTool` to `extends BaseTransactionTool` and remove any local `handleError` override that only wrapped the receipt/precheck error in prose. `BaseTransactionTool` is exported from `@hashgraph/hedera-agent-kit`.

**Related issue(s)**:

Fixes #987

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/packages/core/src/plugins/core-account-query-plugin/tools/queries/get-account-query.ts
+++ b/packages/core/src/plugins/core-account-query-plugin/tools/queries/get-account-query.ts
@@ -1,6 +1,6 @@
 import { z } from 'zod';
 import BigNumber from 'bignumber.js';
-import { Client, Status } from '@hiero-ledger/sdk';
+import { Client } from '@hiero-ledger/sdk';
 import { Context } from '@/shared/configuration';
 import { getMirrornodeService } from '@/shared/hedera-utils/mirrornode/hedera-mirrornode-utils';
 import { BaseTool } from '@/shared/tools';
@@ -74,7 +74,7 @@ export class GetAccountQueryTool extends BaseTool {
       hbarBalance,
     };
     return {
-      raw: { accountId: normalisedParams.accountId, account: accountWithHbar },
+      raw: { accountId: normalisedParams.accountId, account: accountWithHbar, status: 'SUCCESS' },
       humanMessage: postProcess(accountWithHbar),
     };
   }
@@ -93,7 +93,7 @@ export class GetAccountQueryTool extends BaseTool {
     const message = desc + (error instanceof Error ? `: ${error.message}` : '');
     console.error('[get_account_query_tool]', message);
     return {
-      raw: { status: Status.InvalidTransaction, error: message },
+      raw: { status: 'ERROR', error: message },
       humanMessage: message,
     };
   }
--- a/packages/core/src/plugins/core-account-query-plugin/tools/queries/get-account-query.ts
+++ b/packages/core/src/plugins/core-account-query-plugin/tools/queries/get-account-query.ts
@@ -74,7 +74,7 @@ export class GetAccountQueryTool extends BaseTool {
       hbarBalance,
     };
     return {
-      raw: { accountId: normalisedParams.accountId, account: accountWithHbar, status: 'SUCCESS' },
+      raw: { accountId: normalisedParams.accountId, account: accountWithHbar },
       humanMessage: postProcess(accountWithHbar),
     };
   }
@@ -86,16 +86,6 @@ export class GetAccountQueryTool extends BaseTool {
   async secondaryAction(_request: any, _client: Client, _context: Context): Promise<any> {
     // No secondary action for queries
     return null;
-  }
-
-  async handleError(error: unknown, _context: Context): Promise<any> {
-    const desc = 'Failed to get account query';
-    const message = desc + (error instanceof Error ? `: ${error.message}` : '');
-    console.error('[get_account_query_tool]', message);
-    return {
-      raw: { status: 'ERROR', error: message },
-      humanMessage: message,
-    };
   }
 }
 
--- a/packages/core/src/plugins/core-account-query-plugin/tools/queries/get-hbar-balance-query.ts
+++ b/packages/core/src/plugins/core-account-query-plugin/tools/queries/get-hbar-balance-query.ts
@@ -59,7 +59,7 @@ export class GetHbarBalanceQueryTool extends BaseTool {
       normalisedParams.accountId,
     );
     return {
-      raw: { accountId: normalisedParams.accountId, hbarBalance: toHbar(balance).toString() },
+      raw: { accountId: normalisedParams.accountId, hbarBalance: toHbar(balance).toString(), status: 'SUCCESS' },
       humanMessage: postProcess(toHbar(balance).toString() as string, normalisedParams.accountId),
     };
   }
@@ -76,7 +76,7 @@ export class GetHbarBalanceQueryTool extends BaseTool {
     const desc = 'Failed to get HBAR balance';
     const message = desc + (error instanceof Error ? `: ${error.message}` : '');
     console.error('[get_hbar_balance_query_tool]', message);
-    return { raw: { error: message }, humanMessage: message };
+    return { raw: { status: 'ERROR', error: message }, humanMessage: message };
   }
 }
 
--- a/packages/core/src/plugins/core-account-query-plugin/tools/queries/get-hbar-balance-query.ts
+++ b/packages/core/src/plugins/core-account-query-plugin/tools/queries/get-hbar-balance-query.ts
@@ -59,7 +59,7 @@ export class GetHbarBalanceQueryTool extends BaseTool {
       normalisedParams.accountId,
     );
     return {
-      raw: { accountId: normalisedParams.accountId, hbarBalance: toHbar(balance).toString(), status: 'SUCCESS' },
+      raw: { accountId: normalisedParams.accountId, hbarBalance: toHbar(balance).toString() },
       humanMessage: postProcess(toHbar(balance).toString() as string, normalisedParams.accountId),
     };
   }
@@ -70,13 +70,6 @@ export class GetHbarBalanceQueryTool extends BaseTool {
 
   async secondaryAction(request: any, _client: Client, _context: Context) {
     return request;
-  }
-
-  async handleError(error: unknown, _context: Context): Promise<any> {
-    const desc = 'Failed to get HBAR balance';
-    const message = desc + (error instanceof Error ? `: ${error.message}` : '');
-    console.error('[get_hbar_balance_query_tool]', message);
-    return { raw: { status: 'ERROR', error: message }, humanMessage: message };
   }
 }
 
--- a/packages/core/src/shared/base-transaction-tool.ts
+++ b/packages/core/src/shared/base-transaction-tool.ts
@@ -1,34 +1,46 @@
-import { ReceiptStatusError } from '@hiero-ledger/sdk';
+import { PrecheckStatusError, ReceiptStatusError } from '@hiero-ledger/sdk';
 import { BaseTool } from './tools';
 import { Context } from './configuration';
+import { TOOL_STATUS } from '@/shared/utils';
 
 /**
  * Abstract base class for all Hedera transaction tools.
  *
  * Extends {@link BaseTool} with Hedera-specific error handling: when a
- * `ReceiptStatusError` is thrown (i.e. `ExecuteStrategy` received a non-SUCCESS
- * receipt from the network), the SDK status name and transaction ID are lifted
- * into structured fields on `raw` so callers don't need to parse prose.
+ * `ReceiptStatusError` or `PrecheckStatusError` is thrown, the SDK status name
+ * and transaction ID are lifted into structured fields on `raw` so callers
+ * don't need to parse prose.
  *
- * On successful execution `raw.status` is `'SUCCESS'` (set by `ExecuteStrategy`).
- * When a `ReceiptStatusError` is caught by `handleError()`, `raw.status` is
- * `'ERROR'` and the following additional fields are set:
+ * - `ReceiptStatusError` — thrown by `ExecuteStrategy` when `getReceipt()` returns
+ *   a non-SUCCESS status (e.g. `TOKEN_NOT_ASSOCIATED_TO_ACCOUNT`).
+ * - `PrecheckStatusError` — thrown before a transaction reaches the network when
+ *   the node rejects it at the precheck stage (e.g. `INSUFFICIENT_PAYER_BALANCE`).
+ *
+ * Both error types expose `.status` (SDK `Status` enum) and `.transactionId`
+ * (`TransactionId`), so the same structured fields are set for both:
+ * - `raw.status` — always `'ERROR'`
  * - `raw.errorCode` — SDK status name, e.g. `'INSUFFICIENT_PAYER_BALANCE'`
  * - `raw.transactionId` — the transaction ID string
  * - `raw.error` — the original error message
  *
- * This behaviour only applies in `AUTONOMOUS` mode. In
- * `RETURN_BYTES` mode `getReceipt()` is never called, so `ReceiptStatusError`
- * cannot be thrown and `BaseTool.handleError()` handles all errors generically.
+ * On successful execution `raw.status` is `'SUCCESS'` (defaulted by
+ * `BaseTool.execute()` if not set explicitly upstream by `ExecuteStrategy` or
+ * `ReturnBytesStrategy`).
+ *
+ * This behaviour only applies in `AUTONOMOUS` mode. In `RETURN_BYTES` mode
+ * `getReceipt()` is never called, so `ReceiptStatusError` cannot be thrown.
+ * `PrecheckStatusError` can in principle still be thrown in `RETURN_BYTES` mode
+ * if the node rejects the submission itself, but in practice the bytes are
+ * returned before submission and `BaseTool.handleError()` handles all errors.
  */
 export abstract class BaseTransactionTool extends BaseTool {
   async handleError(error: unknown, context: Context): Promise<any> {
-    if (error instanceof ReceiptStatusError) {
+    if (error instanceof ReceiptStatusError || error instanceof PrecheckStatusError) {
       const message = `Failed to execute ${this.name}: ${error.message}`;
       console.error(`[${this.method}]`, message);
       return {
         raw: {
-          status: 'ERROR',
+          status: TOOL_STATUS.ERROR,
           errorCode: error.status.toString(),
           transactionId: error.transactionId.toString(),
           error: error.message,
--- a/packages/core/src/shared/base-transaction-tool.ts
+++ b/packages/core/src/shared/base-transaction-tool.ts
@@ -1,0 +1,39 @@
+import { ReceiptStatusError } from '@hiero-ledger/sdk';
+import { BaseTool } from './tools';
+import { Context } from './configuration';
+
+/**
+ * Abstract base class for all Hedera transaction tools.
+ *
+ * Extends {@link BaseTool} with Hedera-specific error handling: when a
+ * `ReceiptStatusError` is thrown (i.e. `ExecuteStrategy` received a non-SUCCESS
+ * receipt from the network), the SDK status name and transaction ID are lifted
+ * into structured fields on `raw` so callers don't need to parse prose.
+ *
+ * `raw.status` is always `'ERROR'`; the additional fields are:
+ * - `raw.errorCode` — SDK status name, e.g. `'INSUFFICIENT_PAYER_BALANCE'`
+ * - `raw.transactionId` — the transaction ID string
+ * - `raw.error` — the original error message
+ *
+ * This behaviour only applies in `AUTONOMUS` mode. In
+ * `RETURN_BYTES` mode `getReceipt()` is never called, so `ReceiptStatusError`
+ * cannot be thrown and `BaseTool.handleError()` handles all errors generically.
+ */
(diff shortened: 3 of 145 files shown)


packages/core/src/shared/tools.ts
```diff
@@ -20,6 +20,23 @@ export interface Tool {
...
+/**
+ * Base class for all Hedera Agent Kit tools. Subclasses implement the
+ * `coreAction` / `secondaryAction` pipeline and inherit a standard
+ * `{ raw, humanMessage }` output envelope.
+ *
+ * ## `raw.status` contract
+ *
+ * Every tool returns a `raw` object whose `status` field signals outcome:
+ *
+ * | `raw.status`  | Source                                          | Meaning                                                                 |
```
skurzyp-blockydevs: lacks the PARSE_ERROR status

packages/core/src/shared/base-transaction-tool.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Abstract base class for all Hedera transaction tools.
+ *
+ * Extends {@link BaseTool} with Hedera-specific error handling: when a
+ * `ReceiptStatusError` is thrown (i.e. `ExecuteStrategy` received a non-SUCCESS
+ * receipt from the network), the SDK status name and transaction ID are lifted
+ * into structured fields on `raw` so callers don't need to parse prose.
+ *
+ * `raw.status` is always `'ERROR'`; the additional fields are:
```
skurzyp-blockydevs: is always "error" seems to not be the case - tool retuns "SUCCESS" on successful runs too

packages/core/src/shared/base-transaction-tool.ts
```diff
@@ -0,0 +1,41 @@
...
+ * - `raw.transactionId` — the transaction ID string
+ * - `raw.error` — the original error message
+ *
+ * This behaviour only applies in `AUTONOMOUS` mode. In
+ * `RETURN_BYTES` mode `getReceipt()` is never called, so `ReceiptStatusError`
+ * cannot be thrown and `BaseTool.handleError()` handles all errors generically.
+ */
+export abstract class BaseTransactionTool extends BaseTool {
+  async handleError(error: unknown, context: Context): Promise<any> {
+    if (error instanceof ReceiptStatusError) {
```
MWBlocky: This only handles ReceiptStatusError, but PrecheckStatusError also has .status and .transactionId (e.g. INSUFFICIENT_PAYER_BALANCE is usually a precheck error) - please handle it here too, otherwise errorCode and transactionId are lost.

packages/core/src/shared/utils/default-tool-output-parsing.ts
```diff
@@ -123,3 +123,115 @@ export const untypedQueryOutputParser = (rawOutput: string): { raw: any; humanMe
...
+ *   structurally malformed (missing `raw`, non-object, etc.).
+ * - `unknown`: `raw.status` is a non-empty string that does not match any of
+ *   the above (e.g. an unexpected Hedera SDK status from `ExecuteStrategy`).
+ *
+ * See {@link ToolRawStatus} for the known `raw.status` string values and
+ * {@link BaseTool} for the full `raw.status` contract.
+ */
+export type ToolResultStatus<T = unknown> =
+  | { kind: 'success'; transactionId?: string; data: T; humanMessage: string }
+  | { kind: 'failure'; errorCode: number | string; error: string; humanMessage: string }
```
MWBlocky: Please add transactionId?: string to the failure variant - raw has it on receipt failures, but the typed result drops it. Also, errorCode can be just string, the number case never happens.

packages/core/src/plugins/core-account-query-plugin/tools/queries/get-account-query.ts
MWBlocky: This handleError is now the same as the base BaseTool.handleError - can we delete these 10 overrides and just inherit, like the transaction tools do?

packages/core/src/plugins/core-account-query-plugin/tools/queries/get-hbar-balance-query.ts
```diff
@@ -59,7 +59,7 @@ export class GetHbarBalanceQueryTool extends BaseTool {
       normalisedParams.accountId,
     );
     return {
-      raw: { accountId: normalisedParams.accountId, hbarBalance: toHbar(balance).toString() },
+      raw: { accountId: normalisedParams.accountId, hbarBalance: toHbar(balance).toString(), status: 'SUCCESS' },
```
MWBlocky: Instead of adding status: 'SUCCESS' in 10 places, could we set it once in BaseTool.execute (e.g. result.raw.status ??= 'SUCCESS')? Then new tools get it for free.

packages/core/src/shared/utils/default-tool-output-parsing.ts
```diff
@@ -123,3 +123,115 @@ export const untypedQueryOutputParser = (rawOutput: string): { raw: any; humanMe
...
+/**
+ * Well-known string values that appear in `raw.status` across all tool outputs.
+ *
+ * | Value          | Set by                                                                                       |
+ * |----------------|----------------------------------------------------------------------------------------------|
+ * | `'SUCCESS'`    | `ReturnBytesStrategy`, query tool `coreAction` implementations, and `ExecuteStrategy` (receipt is always `SUCCESS` here — non-SUCCESS receipts throw before the return) |
+ * | `'ERROR'`      | `BaseTransactionTool.handleError()` (receipt failures) or `BaseTool.handleError()` (generic). Receipt failures additionally set `raw.errorCode` (e.g. `'INSUFFICIENT_PAYER_BALANCE'`) and `raw.transactionId`. |
+ * | `'PARSE_ERROR'`| `transactionToolOutputParser` / `untypedQueryOutputParser` when the output is malformed      |
+ */
+export type ToolRawStatus = 'SUCCESS' | 'ERROR' | 'PARSE_ERROR';
```
MWBlocky: ToolRawStatus is exported but nothing actually uses it, so a typo like 'SUCESS' still compiles - consider a shared TOOL_STATUS const used at every write site.

MWBlocky:
`packages/langchain/src/responseParserService.ts:9`
This branch returns only { bytes } and drops the new status: 'SUCCESS' field - please use return { ...data, bytes: new Uint8Array(data.bytes.data) } to keep the other fields.

`docs/MIGRATION-v4.md:749`
 This example still uses status: Status.InvalidTransaction - please change it to status: 'ERROR' to match the new contract.

packages/core/src/plugins/core-account-query-plugin/tools/queries/get-hbar-balance-query.ts
skurzyp-blockydevs: done

packages/core/src/shared/utils/default-tool-output-parsing.ts
skurzyp-blockydevs: done

packages/core/src/plugins/core-account-query-plugin/tools/queries/get-account-query.ts
skurzyp-blockydevs: done

MWBlocky (approved):

MWBlocky (approved):

Issue #853, opened 2026-05-23
Kit feedback


KaushalGoud:
### Contact Info

KaushalGoud

### Describe your feedback, bug, or feature request

The Hedera Agent Kit made it much easier to build an AI-powered on-chain workflow with LangChain integration and tool orchestration. The overall developer experience was smooth, especially around `coreQueriesPlugin` and toolkit setup.

One improvement that would help developers is clearer documentation/examples for advanced production patterns such as custom transaction signing flows (AWS KMS/HSM signing), SSE streaming agents, and more complex tool-calling architectures. Additional real-world examples for secure enterprise deployments would also be valuable.

### What type of feedback is this?

- [ ] General Feedback
- [ ] Bug or Suggestion
- [x] Feature Request
- [ ] Other

### What does this feedback apply to?

- [x] Hedera Agent Kit JS
- [ ] Hedera Agent Kit Python
- [ ] Feature Request
- [ ] Other

### Score (1-10): How confident did you feel after reading the docs that you could build successfully?

7

### Score (1-10): How easy was it to get help when you were blocked?

5

### Score (1-10): How intuitive were the APIs / SDKs to use?

8

### Score (1-10): How easy was it to debug issues?

4

### Score (1-10): How likely are you to build again on Hedera again?

10

skurzyp-blockydevs:
@KaushalGoud Hi, thanks for the feedback! We’ll review your suggestions and decide which examples should be included in this repository. We may also choose to link to high-quality third-party guides instead of implementing everything here. If you’ve found any particularly helpful resources during your development, feel free to share them.

Good luck with your project!

skurzyp-blockydevs:
Hey @KaushalGoud! This pending PR addresses the request for custom signing flow in Hedera Agent Kit. This should significantly simplify integration of KMS services. Regarding other suggestions I have create a separate issue for SEE agent (#995) but I'm not sure if we will decide to implement it. This is more agent framework specific problem than related to HAK so developers should look for the guides in the framework docs rather than here.

PR #986, opened 2026-07-06
feat: add configurable transaction strategies and enhance jsdocs

### LINKED ISSUE #886: Feature request: two-party fee delegation primitive (subject ≠ payer signing flow)
## Summary

The current `tx-mode-strategy.ts` models single-party signing only:
- `AgentMode.AUTONOMOUS` → `ExecuteStrategy`: the configured client/agent signs everything
- `AgentMode.RETURN_BYTES` → `ReturnBytesStrategy`: `transactionId = TransactionId.generate(context.accountId)` collapses "signer" and "payer" into one identity

Hedera's native `payerAccountId` field supports a two-party flow where the **subject** (user, whose account state changes) is distinct from the **payer** (agent, who covers HBAR fees on submit). The `RETURN_BYTES` strategy can't construct this — it generates the transaction ID from a single context account, so the bytes are payer = subject by construction.

This issue requests a first-class primitive for two-party signing.

## Use case

We're building a third-party travel-booking platform on Hedera (Xeni). User pays in USDC; we want to cover the HBAR transaction fee (~$0.001) so end users never need to hold HBAR.

The required transaction shape:
- `TransferTransaction` (user → treasury, USDC amount)
- `transactionId = TransactionId.generate(agentAccountId)` so the agent is the named payer
- User signature attached (user is the asset owner, must sign)
- Agent signature attached as the named payer

Both signatures are required by Hedera protocol. The current strategies can't model this without us building it externally — which we did (see "Reference implementation" below), but other teams hitting this pattern will reinvent the same thing.

### Option A — New `AgentMode.DELEGATED_PAYER` mode

Strategy variant that:
- Accepts a `subjectAccountId` separately from `payerAccountId` in context
- Constructs the transaction with `setTransactionId(TransactionId.generate(payerAccountId))`
- Returns bytes ready for the subject to sign via wallet (HashPack, etc.)

Minimal change to the existing strategy interface; backward-compatible. The submit-side merging stays as a separate user-land concern.

### Option B — New upstream tool `submit_with_added_signature`

A tool that:
- Accepts user-signed bytes + a list of additional signers to apply
- Validates `payerAccountId === expectedPayerAccountId`
- Adds the configured signatures (`tx.addSignature(payerKey, sig)` or `tx.sign(payerPrivateKey)`)
- Submits via the standard execute → getReceipt pipeline

This is what we built locally and would happily contribute upstream. Tool name TBD by the maintainers; the shape is the contribution.

### Option C — Multi-party signing primitives in the strategy interface (the right architectural shape)

Model "subject account" vs "payer account" as distinct context fields. The strategy is then responsible for the right `TransactionId.generate(payerAccountId)` + signature collection orchestration. AUTONOMOUS / RETURN_BYTES / DELEGATED_PAYER all become specializations of the same surface.

Larger change but the most principled — the single-party assumption is baked into the current strategy interface, and a multi-party-first model would compose cleanly with future use cases (multi-sig treasuries, delegated allowance flows, etc.).

## Reference implementation

We built this in our own repo (since waiting on upstream was a launch-blocking dep). The relevant files:

- [`src/plugins/xeniWrite/submitWithPayerSignature.ts`](https://github.com/xeni-app/hedera-xeni-mcp-server/blob/develop/src/plugins/xeniWrite/submitWithPayerSignature.ts) — accepts pre-merged signed bytes, validates `payerAccountId === agentAccountId`, adds the agent signature, submits
- [`src/plugins/xeniWrite/submitUserSigMapWithPayerSignature.ts`](https://github.com/xeni-app/hedera-xeni-mcp-server/blob/develop/src/plugins/xeniWrite/submitUserSigMapWithPayerSignature.ts) — sibling that accepts unsigned tx bytes + a SignatureMap protobuf separately (cleaner FE contract — wallet never has to materialize a `Transaction` instance)
- [`src/plugins/xeniWrite/buildAgentPayerTransferTx.ts`](https://github.com/xeni-app/hedera-xeni-mcp-server/blob/develop/src/plugins/xeniWrite/buildAgentPayerTransferTx.ts) — pre-builds the unsigned-but-frozen tx with `setNodeAccountIds([new AccountId(3)])` for deterministic-bytes-for-wallet-signing
- [`src/plugins/xeniWrite/buildAgentPayerTokenAssociateTx.ts`](https://github.com/xeni-app/hedera-xeni-mcp-server/blob/develop/src/plugins/xeniWrite/buildAgentPayerTokenAssociateTx.ts) — sibling for HTS associations

Battle-tested with 40+ unit tests including drift-guards on tx-type allowlists + defense-in-depth signature-count assertions catching silent `tx.sign` / `tx.addSignature` no-ops.

Happy to contribute these upstream if the design lands in a form that matches the project's preferences — let us know which Option (A/B/C) you'd want us to align on and we can open a PR.

## What this enables (and what it avoids)

**Enables**: any agent / relayer / dApp that wants users to pay in a token (USDC, HTS-native or otherwise) while a service covers the gas — without building bespoke two-party signing infrastructure.

**Avoids**: each team that hits this pattern reinventing the same `Transaction.fromBytes → addSignature → sign → execute` pipeline + the same validation primitives (`payerAccountId === expectedPayer`, tx-type allowlist, signature-count defense-in-depth).

## References

- Hedera SDK [`Transaction.addSignature`](https://github.com/hiero-ledger/hiero-sdk-js) — the underlying primitive
- Our spec doc explaining the use case at the protocol layer: [`USDC_DESIGN.md §8.1`](https://github.com/xeni-app/hedera-xeni-mcp-server/blob/develop/docs/USDC_DESIGN.md#81-hashpack-native-hedera) (Hedera-native fee delegation)
- Related docs request (separate issue, to be filed): "USDC on Hedera EIP-2612 status clarification in Agent Kit docs" — since `permit`-based fee delegation is the standard EVM workaround and is NOT supported on Hedera USDC (verified on-chain testnet + mainnet), the docs should explicitly tell users not to go down that path

Thanks for considering!

### LINKED ISSUE #985: feat: allow implementing custom signing strategies supporting remote signers
This issue is fromed based on:
- general feedback regarding Ai bounties
- insights gathered while reviewing the bounties

Currently the hedera agent kit is limited to two modes: `AgentMode.AUTONOMUS` or `AgentMode.RETURN_BYTES`. There is no way to implement a custom signig strategi which forces users to use `AgentMode.RETURN_BYTES` in multiple flows and add the implementation layer post tool exectution rather than embed the logic withing tool lifecycle.

By adding option to create custom strategies developers will be able to intercept the tool execution with external calls to remote signers (TEE, MPC) or implement HITL modes right there instead of after the tool execution is finished by intercepting the agent ReAct loop.

skurzyp-blockydevs:
**Description**:

**Related issue(s)**:

Fixes #985 
Fixes #886 
Refferences #853 

**Notes for reviewer**:

**Checklist**


packages/core/src/shared/api.ts
```diff
@@ -16,6 +17,15 @@ class HederaAgentAPI {
       throw new Error('Client must be connected to a network');
     }
     this.context = context || {};
+    
+    if (this.context.mode === AgentMode.RETURN_BYTES && !this.context.accountId) {
+      throw new Error('accountId must be provided in Context when AgentMode is RETURN_BYTES');
+    }
```
kierzniak: The new guard throws in the toolkit constructor, so read-only / pre-wallet setups that used to build fine now 500 even for pure read queries.

packages/core/src/shared/strategies/http-signing-strategy.ts
kierzniak, line 1: Does this strategy is compatible with some existing service? Maybe it should be included in examples? From what I can see AWS KMS or GCP KMS works differently.

packages/core/src/shared/configuration.ts
```diff
@@ -1,34 +1,96 @@
...
   mode?: AgentMode;
 
-  // Mirrornode config
+  /**
+   * A custom transaction strategy class or function that handles signing and broadcasting.
+   * Required when `mode` is set to `AgentMode.CUSTOM`. Note that most custom strategies
+   * (like the built-in `HttpSigningStrategy`) also require `accountId` to be set in the Context
+   * to determine the transaction payer.
+   */
+  transactionStrategy?: TxModeStrategy;
```
kierzniak: JSDoc lies about the type. Says transactionStrategy accepts a "class or function", but the type requires an instance with handle().

kierzniak (changes requested):

packages/core/src/shared/strategies/http-signing-strategy.ts
skurzyp-blockydevs: ill remove it

docs/HOOKS_AND_POLICIES.md
```diff
@@ -84,8 +84,12 @@ const context = {
 Provides an immutable audit trail by logging tool executions to a Hedera Consensus Service (HCS) topic.
 
 > [!IMPORTANT]  
-> **Autonomous Mode Only**: This hook is strictly available in `AUTONOMOUS` mode. It will throw an error if used in
-> `RETURN_BYTES` mode.
+> **Autonomous Mode Only**: This hook only supports `AgentMode.AUTONOMOUS`. It throws before the tool executes in:
```
skurzyp-blockydevs: Added the CUSTOM mode to not supported by out audit trail hooks because we dont know what this strategy is returning. Users would need to implement their own approach to creating a methods for parsing output and posting them to the HCS.

docs/HOOKS_AND_POLICIES.md
```diff
@@ -154,17 +162,27 @@ INDEXED registry as the session topic to list audit entries.
 
 - `relevantTools`: `string[]` - List of tool names that trigger audit trail logging.
 - `sessionId`: `string` - The Hedera topic ID (format `0.0.xxx`) used as the audit session registry.
+- `loggingClient?`: `Client` - (Optional) A separate Hedera client used to submit audit messages to HCS. Defaults to
+  the agent's operator client. Must have a local operator key capable of signing and submitting transactions.
```
skurzyp-blockydevs, line 172: The HCS audit trail was extended with an option to pass a custom loging client - this means the agent can be ran using one account but the audit is performed using another one. The HOL strategy was missing this feature

docs/TRANSACTION_MODES.md
```diff
@@ -81,17 +81,17 @@ This is the recommended approach for integrating secure and institutional key-ma
...
+> In `CUSTOM` mode, passing `accountId` in the `Context` is highly recommended. Most custom strategies require it to set the correct payer on the transaction before sending it to the remote service.
 
 ### Defining a Custom Strategy (Reference Implementation)
 
 Below is a reference implementation of a custom strategy using a remote enclave/TEE endpoint:
 
 ```typescript
 import { Client, Transaction } from '@hiero-ledger/sdk';
-import { TxModeStrategy, RawTransactionResponse, Context } from '@hashgraph/hedera-agent-kit';
+import { TransactionStrategy, RawTransactionResponse, Context } from '@hashgraph/hedera-agent-kit';
```
skurzyp-blockydevs, line 110: @kierzniak I have renamed the TxModeStrategy to TransactionStrategy. What else do you suggest to rename?

packages/core/src/shared/configuration.ts
skurzyp-blockydevs: fixed

packages/core/src/shared/api.ts
skurzyp-blockydevs: removed

kierzniak (dismissed):

skurzyp-blockydevs:
Decided to add two custom modes: one for custom tx execution, other for custom bytes return. This aligns with a feature request #886

packages/core/src/shared/hedera-utils/mirrornode/hedera-mirrornode-service-default-impl.ts
```diff
@@ -32,23 +32,33 @@ export class HederaMirrornodeServiceDefaultImpl implements IHederaMirrornodeServ
...
-   * write would otherwise fail. Non-404 statuses throw immediately; an
-   * exhausted 404 throws via `buildError`.
+   * Fetches JSON from the mirror node with exponential backoff retry.
+   * Retries on 404 (entity not yet indexed), 429, 502, 503, 504 (transient
+   * overload), and network-level errors (fetch throws). Non-retryable statuses
+   * throw immediately via `buildError`.
    */
   private async fetchJson<T>(url: string, buildError: (response: Response) => string): Promise<T> {
-    const maxAttempts = 3;
+    const RETRYABLE_STATUSES = new Set([404, 429, 502, 503, 504]);
```
MWBlocky, line 41: 404 is retryable and maxAttempts went 3→5, so a genuinely non-existent entity now blocks ~15s (1+2+4+8) before failing, vs ~3s before. 404 conflates "not yet indexed" with "does not exist"; every legitimate not-found query pays the full backoff. Consider a lower cap for 404 specifically, or a total-wait ceiling.

MWBlocky (commented):
docs/TRANSACTION_MODES.md, packages/core/README.md, examples/nextjs/README.md (many lines) — Docs call the interface TxModeStrategy throughout, but it was renamed to TransactionStrategy<TResult> (the ADR itself records the rename). TxModeStrategy is not exported — anyone copy-pasting will look for a symbol that doesn't exist. Global find/replace
  TxModeStrategy → TransactionStrategy.

packages/core/src/shared/hedera-utils/mirrornode/hedera-mirrornode-service-default-impl.ts
skurzyp-blockydevs: i think that there is some problem with the mirror node in the hiero solo node - those errors are not throwing on the testnet and there we dont need to wait that long for the smartcontract info

MWBlocky (approved):

Issue #854, opened 2026-05-23
HBAR Arcade — https://github.com/Lakshmikanth-3/Arcade


Lakshmikanth-3:
Used HederaAgentAPI from @hashgraph/hedera-agent-kit v4 to build a 
pay-to-play AI arcade: entry fee TX → Groq LLM question generation → 
answer evaluation → prize payout TX → HCS game log.
Feedback:
1. HederaAgentKit class is not exported in v4 — only HederaAgentAPI exists. 
   README still references the old class name, causing setup confusion.
2. HederaAgentAPI constructor parameters are undocumented — had to read 
   TypeScript types directly.
3. No example for non-OpenAI LLMs (Groq, Anthropic) — would help builders 
   using free-tier models via LangChain.
4. Error "Cannot read properties of undefined (reading 'ledgerId')" when 
   calling HederaAgentAPI without args — a friendlier message would save time.
Overall great kit — the Zod tool schemas (transferHbarParameters, 
submitTopicMessageParameters) were especially useful for validation.

skurzyp-blockydevs:
Hi @Lakshmikanth-3! Thank you for your feedback.

We appreciate you catching the outdated README reference and the undocumented parameters of the `HederaAgentAPI` constructor - we’ll definitely fix those. We’ll also discuss internally how to address the other reported pain points.

Please note that implementing these changes may take some time, as we have limited resources. In the meantime, feel free to open a PR with any fixes.

Good luck with your project!

skurzyp-blockydevs:
Hey @Lakshmikanth-3, the `HederaAgentAPI` is not intended for a direct calling. It is an internal class for building framework adapters. Migration to v4 change the architecture of the SKD. Now we have a core SDK class and separate adapters for other frameworks. We noticed that this was not clearly addressed in the main package readme. PRs  #990 and $989 fix that. 
We added better JS docs for `HederaAgentAPI` in https://github.com/hashgraph/hedera-agent-kit-js/pull/992

skurzyp-blockydevs:
Regarding "No example for non-OpenAI LLMs (Groq, Anthropic) — would help builders
using free-tier models via LangChain." - this is out of scope for this sdk. Developers should follow guide of frameworks that they have decided to use.

PR #838, opened 2026-05-21
feat: expose agent tool summaries

Adamchaua:
Privacy cleanup: removed public payment/private-contact details. Technical context intentionally omitted from this cleanup.

PR #992, opened 2026-07-07
feat: expose method for listing registered tools in the`HederaAgentAPI`

### LINKED ISSUE #834: HAK v4: Plugin registration order affects tool availability in HederaAgentAPI
## Feedback from Hedera AI Agent Bounty — Week 1

**Project:** Hedera AI Court (https://github.com/jmgomezl/hedera-ai-court)

### Summary
When building with HAK v4 and the plain-object plugin pattern, I noticed that tool availability in `HederaAgentAPI` depends on plugin registration happening before the agent processes its first message. Adding clearer documentation or a startup validation log (e.g. "Registered tools: [...]") would help developers verify their plugins loaded correctly.

### Steps to reproduce
1. Instantiate `HederaAgentAPI` with a custom plugin
2. Call `agent.chat()` before verifying tool registration
3. Tools may silently not be available if plugin was registered incorrectly

### Suggested improvement
Add a `agent.listTools()` method or startup log that lists all registered tools by name so developers can confirm plugin registration succeeded.

### Environment
- `@hashgraph/hedera-agent-kit` v4
- Node.js 20, TypeScript, ESM

skurzyp-blockydevs:
**Description**:

**Related issue(s)**:

Fixes #834 , #854

**Notes for reviewer**:

**Checklist**


kierzniak (dismissed):
Hmm why we need this? I think we have method to get all registered tools.

skurzyp-blockydevs:
HederaAgentAPI is not exposing any getter for getting the list of registered tools. I think this might be convenient for devs building other adapters than those that are included in the HAK. It was requested and a PR with initial soultion was created here: #838.

packages/core/src/shared/api.ts
```diff
@@ -1,15 +1,82 @@
...
   tools: Tool[];
 
+  /**
+   * @param client - Connected Hiero SDK `Client`; must have `ledgerId` set.
+   * @param context - Optional execution context; defaults to `{}`.
+   * @param tools - Optional tool registry; defaults to `[]`.
+   * @throws {Error} When `client.ledgerId` is falsy ("Client must be connected to a network").
+   */
   constructor(client: Client, context?: Context, tools?: Tool[]) {
     this.client = client;
```
MWBlocky, line 83: add guards for:
```
client === undefined
this.client.ledgerId ledgerId could be undefined too
```

MWBlocky: consider add:
```
(if (!client) throw new Error('HederaAgentAPI requires a connected Client')
```

MWBlocky:
Small correction: the tool list is already reachable - api.tools is public. The real point of listTools() is that it returns only method, name and description, without the execute handlers and Zod schemas. So the result is safe to log or serialize (e.g. "Registered tools: [...]" at startup), which is exactly what the bounty feedback asked for.

MWBlocky (dismissed):

MWBlocky (approved):

PR #1011, opened 2026-07-15
fix: erc20 display units

### LINKED ISSUE #298: Unify parameters format for evm tools
Some tools take value in smallest denoms and other in display units (f.e. ERC20 initial supply) and multiply them to the correct value on the contract level.

One common approach should be created. Similarity to the HTS operations would probably be best otherwise the differences should be highlighted in the docs

MWBlocky:
**Related issue(s)**:

Fixes #298

### DIFF
--- a/packages/core/src/shared/hedera-utils/decimals-utils.ts
+++ b/packages/core/src/shared/hedera-utils/decimals-utils.ts
@@ -1,4 +1,35 @@
 import BigNumber from 'bignumber.js';
+import { IHederaMirrornodeService } from './mirrornode/hedera-mirrornode-service.interface';
+
+// Function selector of ERC20 `decimals()`
+const ERC20_DECIMALS_SELECTOR = '0x313ce567';
+
+/**
+ * Reads the `decimals()` value of an ERC20 contract via the mirror node
+ * read-only `/contracts/call` endpoint
+ *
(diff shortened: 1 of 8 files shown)


skurzyp-blockydevs (approved):

packages/core/src/shared/hedera-utils/hedera-parameter-normaliser.ts
skurzyp-blockydevs, line 968: does this method accept display units? You changed the description but not the code. Shouldnt it cast the supply to base units here?

packages/core/src/shared/hedera-utils/decimals-utils.ts
```diff
@@ -1,4 +1,35 @@
...
+    body: JSON.stringify({ data: ERC20_DECIMALS_SELECTOR, to: contractInfo.evm_address }),
+  });
+  if (!response.ok) {
+    throw new Error(
+      `Failed to read decimals of ERC20 contract ${contractId}: ${response.status} ${response.statusText}`,
+    );
+  }
+  const { result } = (await response.json()) as { result: string };
+  return Number(BigInt(result));
+}
```
skurzyp-blockydevs, line 32: I think this implementation should be in the mirronode service impl. It will then benefit from the retry/back off logic. Here we can just call the mirronode service. What do you think?

PR #1012, opened 2026-07-15
feat: return transaction context alongside bytes in RETURN_BYTES mode

### LINKED ISSUE #926: DX feedback from building an HBAR-payment dapp on top of hedera-agent-kit
# DX feedback from building an HBAR-payment dapp on top of hedera-agent-kit

> Repo this is filed against: https://github.com/hashgraph/hedera-agent-kit-js
> Source: built [hbario](https://hbario.com) — an AI workflow agent that runs
> Hedera Agent Kit in `RETURN_BYTES` mode and hands tx bytes off to HashPack
> via WalletConnect. Six concrete papercuts, ordered by impact.

---

## 1. `RETURN_BYTES` mode loses transaction context needed for downstream signing  *(biggest impact)*

In `AgentMode.RETURN_BYTES`, write tools (HBAR transfer, token mint, topic submit, …) return effectively `{ bytes: Uint8Array }` and nothing else. To hand off cleanly to a wallet (WalletConnect / HashPack) the caller still needs:

- the transaction's expected payer
- a stable transaction ID (for memos, deeplinks, post-execution verification)
- a typed summary of what the tx will do (so the UI can show "Send 1.0 HBAR to 0.0.1234" *before* the wallet popup)
- expected amount / recipient for mirror-node verification after submit

Today we recover that by either re-parsing with `Transaction.fromBytes` (union type, narrowing per tool) or by shadowing the tool inputs in our own wrapper layer. Both are brittle.

**Ask:** Return a structured envelope alongside `bytes`, e.g.

```ts
{
  bytes: Uint8Array;
  transactionId: string;
  type: 'CRYPTO_TRANSFER' | 'TOKEN_MINT' | …;
  summary: { payer: string; transfers: Array<{ accountId: string; amount: string }>; memo?: string };
}
```

This single change would delete ~200 lines of parsing/shadowing code in our app and is the largest gap between "kit gave me bytes" and "wallet signed something I can verify."

## 2. No first-class WalletConnect / HITL example

Every example in the README and `examples/` uses an operator key on the server. The HITL story (`RETURN_BYTES`) is one bullet. For any production dapp that doesn't custody keys, the wallet handoff is left as an exercise.

**Ask:** Add a reference example showing `RETURN_BYTES` + `@hashgraph/hedera-wallet-connect` end-to-end (Next.js or plain Node, ~100 lines). This is the most common "how do I actually use this in a real dapp" question.

## 3. Plugin imports are too granular and inconsistently named

To get a baseline toolkit you import eight plugins:

```ts
import {
  coreAccountPlugin, coreAccountQueryPlugin,
  coreConsensusPlugin, coreConsensusQueryPlugin,
  coreMiscQueriesPlugin,        // ← "Queries" (plural)
  coreTokenPlugin, coreTokenQueryPlugin,
  coreTransactionQueryPlugin,   // ← "Query" (singular)
} from "hedera-agent-kit";
```

- Naming is inconsistent (`coreMiscQueriesPlugin` vs `coreXxxQueryPlugin`).
- There is no convenience bundle.

**Ask:** Ship `coreReadOnlyPlugins`, `coreWritePlugins`, and `corePlugins` arrays, and normalize naming.

## 4. Toolkit isn't network-aware after construction

The `Client` is passed in at construction time, so to support both testnet and mainnet we cache toolkits per `(network, accountId)`. If a user switches network in HashPack mid-session we have to evict the cached toolkit and rebuild it.

**Ask:** Either take `network: 'testnet' | 'mainnet'` as config (kit owns client construction/caching), or expose `toolkit.withNetwork('mainnet')` that returns a cheaply-cloned instance.

## 5. Type ergonomics

- `HederaLangchainToolkit#getTools()` returns `StructuredTool[]`, but the underlying zod input schemas aren't exported. Direct callers (we expose an MCP server) have to introspect `tool.schema` at runtime instead of importing typed schemas.
- `AgentMode` values (`RETURN_BYTES`, `AUTONOMOUS`, …) have no JSDoc in the published `.d.ts`. A one-line description of behavior on each variant would help IDE hover discovery.

## 6. Mirror-node verification is left entirely to the caller

After the wallet signs and submits, you still need to confirm the tx landed on consensus. We wrote our own `verifyTransactionOnMirror(network, txId, expected)` against the REST API. This is universal enough that the kit could ship:

```ts
await toolkit.waitForReceipt(transactionId);     // resolves on consensus
await toolkit.verifyTransfer(transactionId, expected);
```

Or, minimally, expose `toolkit.mirrorNodeBaseUrl` so callers don't have to hardcode mainnet/testnet hostnames.

---

### Context

- App: https://hbario.com  (Render-hosted Next.js, Postgres)
- Repo: https://github.com/mdoty4/hbario
- Integration seam: [`src/lib/hedera/agentKit.ts`](https://github.com/mdoty4/hbario/blob/main/src/lib/hedera/agentKit.ts)
- Mirror-node verification helper we ended up writing: [`src/lib/hedera/mirrorNode.ts`](https://github.com/mdoty4/hbario/blob/main/src/lib/hedera/mirrorNode.ts)

Happy to PR any of the smaller items (3, 5, plugin bundles + JSDoc) if it'd help.

MWBlocky:
**Description**:

  ## Why

  In `RETURN_BYTES` mode, write tools returned only `{ bytes }`. Apps handing
  those bytes to a wallet (WalletConnect / HashPack) still need the transaction
  ID for mirror-node verification, the expected payer, and a human-readable type
  for their own UI — today they recover all of that by re-parsing with
  `Transaction.fromBytes` or by shadowing tool inputs in a wrapper layer.
  The kit already computes this data and throws it away.

  Addresses point 1 of the DX feedback from hbario.com (RETURN_BYTES mode loses
  transaction context needed for downstream signing).

  ## What

  - `RETURN_BYTES` tools now return a `ReturnBytesResult` envelope:
    `bytes`, `transactionId`, `payerAccountId`, `type` (SDK class name),
    `expiresAt` (ISO, `validStart` + `transactionValidDuration`) and `memo`.
    Purely additive — existing consumers of `bytes` are unaffected.
  - `ReturnBytesResult` is exported from the package root; `AgentMode` variants
    got JSDoc describing the behavior of each mode.
  - langchain adapter: `convertBytes` no longer strips the extra envelope
    fields when converting serialized `Buffer` bytes back to `Uint8Array`.
  - nextjs example: the chat message for a pending transaction now shows the
    full envelope (type, payer, id, memo, expiry); wallet rejection is final
    (no re-prompt loop) and non-`Error` rejections render their message
    instead of `[object Object]`.

  ## Notes

  - `type` is derived from `_getLogId()` because the SDK's ESM build is
    minified, which makes `constructor.name` unusable (`"T"`).

**Related issue(s)**:

Fixes #926


packages/core/src/shared/strategies/tx-mode-strategy.ts
```diff
@@ -63,13 +63,43 @@ export class ExecuteStrategy implements TxModeStrategy {
...
       throw new Error('Account ID is required in context for RETURN_BYTES mode');
     const id = TransactionId.generate(context.accountId);
     tx.setTransactionId(id).freezeWith(client);
-    return { bytes: tx.toBytes() };
+    return {
+      bytes: tx.toBytes(),
+      transactionId: id.toString(),
+      payerAccountId: context.accountId,
+      // The SDK's ESM build is minified, so constructor.name is unusable;
+      // _getLogId() embeds the class name as a string literal ("TransferTransaction:<ts>").
```
skurzyp-blockydevs: Are we sure that this will work for all type of transactions places by the tools? Maybe we should extend the coverage in `packages/core/tests/unit/tx-mode-strategy.unit.test.ts`?

packages/core/src/shared/strategies/tx-mode-strategy.ts
```diff
@@ -63,13 +63,43 @@ export class ExecuteStrategy implements TxModeStrategy {
   }
 }
 
+export interface ReturnBytesResult {
```
skurzyp-blockydevs, line 97: Since we are starting to type the responses in a more detailed and structured way we should probably improve our documentation.

Also, I think that this PR should wait and be merged after https://github.com/hashgraph/hedera-agent-kit-js/pull/983

packages/core/src/shared/strategies/tx-mode-strategy.ts
skurzyp-blockydevs: Also, this approach seems to be risky and can start failing in the future. `_getLogId()` is a private method and not a part of public API.

skurzyp-blockydevs (changes requested):

packages/core/src/shared/strategies/tx-mode-strategy.ts
skurzyp-blockydevs: I thought about improvements in the `docs` directory. The docs there never mention the typed responses. Also, this PR introduces the changes to the nextjs example but we should incorporate all those changes to the langchain/adk/ai-sdk and others in `/examples`. 

What I would expect as a dev using the SDK is to see the whole information returned in the return bytes mode visible in the console in the return bytes examples for those frameworks.

skurzyp-blockydevs (changes requested):

examples/adk/return-bytes-agent.ts
```diff
@@ -0,0 +1,90 @@
...
+function validateEnv() {
+  if (!process.env.ACCOUNT_ID) {
+    console.error('Missing required environment variable: ACCOUNT_ID');
+    console.error('Copy .env.example to .env and fill it in.');
+    process.exit(1);
+  }
+}
+
+// After the JSON round-trip through the ADK adapter, `bytes` is a numeric-keyed object, not a
+// Uint8Array — rebuild it.
```
skurzyp-blockydevs: Why is that? I thought that we are returning always Uint8Array because it is not changed during the jsonStringify->jsonParse. The problem was with a buffer - especially in browser

examples/langchain-v1/return-bytes-tool-calling-agent.ts
```diff
@@ -123,8 +123,28 @@ async function bootstrap(): Promise<void> {
...
+            },
+            null,
+            2,
+          ),
+        );
+
         console.log('Transaction bytes found. Executing...');
-        const bytes = toolCall.parsedData.raw.bytes;
+        const bytes = envelope.bytes;
         const tx = Transaction.fromBytes(bytes);
```
skurzyp-blockydevs, line 148: And here we dont cast the bytes back to Uint8Array

skurzyp-blockydevs (changes requested):

skurzyp-blockydevs (approved):

Issue #911, opened 2026-06-08
Agent Kit Feedback:Feedback from building a payment-gated commerce agent with Hedera Agent Kit


Amthebest14:
### Contact Info

@Amthebest14

### Describe your feedback, bug, or feature request

Building Hashpilot — A Payment-Gated AI Commerce Agent with Hedera Agent Kit

We integrated the Hedera Agent Kit into Hashpilot (https://hashpilot.vercel.app/), an AI-powered commerce agent that sells premium financial intelligence services gated behind live HTS token payments in HBAR, USDC, and SAUCE.

What we liked: The Agent Kit significantly accelerated our Hedera-native transaction construction. Having a structured framework for initialising the Hedera client, managing operator credentials, and submitting transactions made the backend setup cleaner than working with raw @hashgraph/sdk alone. The overall architecture encourages proper patterns for treasury-style operations which matched our use case well.

What we ran into:

ESM import resolution: Running the kit in a Vercel serverless environment required explicit .js extensions on all local imports. This caused HTTP 500 failures that were difficult to diagnose initially. Clearer documentation on ESM-specific deployment (especially for Vercel/Node ESM environments) would save builders significant debugging time.
Multi-token TransferTransaction: When we needed to support dynamic HTS token selection at runtime (user chooses HBAR, USDC, or SAUCE at payment time), there were no high-level helper methods for multi-asset routing. We worked around this with direct @hashgraph/sdk TransferTransaction calls, but built-in support for multi-token payment flows would be very useful for commerce-focused agent builders.
ContractUpdateTransaction: Configuring maxAutomaticTokenAssociations on our deployed vault contract required dropping down to raw SDK calls. A Kit-level abstraction for common contract admin operations would improve the experience.
What we'd love to see:

A dedicated commerce/payment-gating example in the docs showing AP2-style gated capability flows
Helper utilities for multi-asset HTS payment routing
Vercel/serverless ESM deployment guide in the official documentation
Overall a solid foundation—the kit gave us the right building blocks to ship a working commerce agent on Hedera Testnet. Happy to elaborate on any of the above.

Project: https://github.com/Amthebest14/hashpilot

### What type of feedback is this?

- [x] General Feedback
- [ ] Bug or Suggestion
- [ ] Feature Request
- [ ] Other

### What does this feedback apply to?

- [x] Hedera Agent Kit JS
- [x] Hedera Agent Kit Python
- [ ] Feature Request
- [ ] Other

[...]

PR #1020, opened 2026-07-23
feat: enforce HTS int64 limits and enhance errors for token allowances

skurzyp-blockydevs:
- Implemented int64 overflow guards for HTS token allowances.
- Enriched error messages with actionable hints for TOKEN_NOT_ASSOCIATED_TO_ACCOUNT.
- Updated tests and documentation to reflect new validations.

**Description**:

**Related issue(s)**:

Related to #908 

**Notes for reviewer**:

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

PR #1024, opened 2026-07-23
feat: transfer fungible token tool

skurzyp-blockydevs:
**Description**:

**Related issue(s)**:

related to #911 

**Notes for reviewer**:

**Checklist**

- [x] Documented (Code comments, README, etc.)
- [x] Tested (unit, integration, etc.)


packages/core/src/shared/hedera-utils/hedera-parameter-normaliser.ts
```diff
@@ -612,6 +614,58 @@ export default class HederaParameterNormaliser {
...
+      client,
+    );
+
+    const tokenInfo = await mirrornode.getTokenInfo(parsedParams.tokenId);
+    const tokenDecimals = Number(tokenInfo.decimals);
+
+    const tokenTransfers: TokenTransferMinimalParams[] = [];
+    let totalAmount = 0;
+
+    for (const transfer of parsedParams.transfers) {
```
MWBlocky, line 646: ### Token transfers are not guaranteed to net to zero (transaction rejected on-chain)

In `normaliseTransferFungibleToken`, the sender's debit is computed **independently** from the recipients' credits:

```ts
let totalAmount = 0;
for (const transfer of parsedParams.transfers) {
  totalAmount += transfer.amount;                                  // sum in JS float, display units
  tokenTransfers.push({ ..., amount: toBaseUnit(transfer.amount, tokenDecimals).toNumber() });
}
tokenTransfers.push({ ..., accountId: senderAccountId,
  amount: toBaseUnit(-totalAmount, tokenDecimals).toNumber() });   // line 653 — floored separately
```

Hedera requires the token transfer list to sum to exactly zero per token; a non-zero sum is rejected at precheck (`INVALID_ACCOUNT_AMOUNTS`). Because the debit is derived by re-summing the display-unit amounts in JS floating point and flooring that total **separately** from the per-recipient floors, the credits and the debit do not always cancel out.

**Reproduced (real `toBaseUnit`, not mocked):**

| Input | Resulting base units | Sum |
|---|---|---|
| `0.1` + `0.2` @ 8 decimals | `[10000000, 20000000, -30000001]` | **−1** ❌ |
| `1.5` @ 0 decimals | `[1, -2]` | **−1** ❌ |
| `0.001` ×2 @ 2 decimals | `[0, 0, -1]` | **−1** ❌ |

**Root cause:** two combined effects — (1) floating-point accumulation, e.g. `0.1 + 0.2 === 0.30000000000000004`, and (2) floor asymmetry, `floor(-x) !== -floor(x)` for non-integer `x`. `0.1`/`0.2` to two recipients on an 8-decimal token is an ordinary transfer, so this hits the common path.

**Why tests don't catch it:** `transfer-fungible-token-params-normalization.unit.test.ts` mocks `toBaseUnit` as `amount * 10 ** decimals` (no flooring) and only uses integer amounts, so the imbalance can never appear.

**Suggested fix** — derive the sender's debit from the sum of the already-converted base-unit credits, so the list nets to zero by construction:

```ts
const safeDecimals = Number.isFinite(tokenDecimals) ? tokenDecimals : 0;
let totalBase = 0;
for (const transfer of parsedParams.transfers) {
  const base = toBaseUnit(transfer.amount, safeDecimals).toNumber();
  totalBase += base;
  tokenTransfers.push({ tokenId: parsedParams.tokenId, accountId: transfer.accountId, amount: base });
}
tokenTransfers.push({ tokenId: parsedParams.tokenId, accountId: senderAccountId, amount: -totalBase });
```

Please also add a regression test that uses the **real** `toBaseUnit` (not a mock) and asserts the transfer list sums to zero for fractional inputs such as `0.1` + `0.2`.

> Note: the same pattern exists in `normaliseTransferFungibleTokenWithAllowance` (line ~687, `toBaseUnit(-totalAmount, ...)`), which has the identical defect and is worth fixing together.

packages/core/src/shared/hedera-utils/hedera-parameter-normaliser.ts
```diff
@@ -615,6 +617,63 @@ export default class HederaParameterNormaliser {
...
+    const tokenInfo = await mirrornode.getTokenInfo(parsedParams.tokenId);
+    const tokenDecimals = Number(tokenInfo.decimals);
+    // Fallback to 0 if decimals are missing or NaN
+    const safeDecimals = Number.isFinite(tokenDecimals) ? tokenDecimals : 0;
+
+    const tokenTransfers: TokenTransferMinimalParams[] = [];
+    let totalBase = 0;
+
+    for (const transfer of parsedParams.transfers) {
+      const base = toBaseUnit(transfer.amount, safeDecimals).toNumber();
```
kierzniak: Maybe we should use toBaseUnitLong(transfer.amount, safeDecimals, 'Transfer amount') here, like every sibling? .toNumber() drops to float64, so we lose the int64 guard #1020 just added

docs/HEDERATOOLS.md
```diff
@@ -836,6 +837,41 @@ Remove allowance for spender 0.0.123 on tokens 0.0.456 and 0.0.789
 
 ---
 
+### TRANSFER_FUNGIBLE_TOKEN_TOOL
+
+**Supports Hooks & Policies**: ✅ Yes
```
kierzniak, line 842: Not true yet for policies. `TRANSFER_FUNGIBLE_TOKEN_TOOL` is missing from both `relevantTools (max-recipients-policy.ts:50)` and `_getBuiltinStrategies()` `(:138)`, so `MaxRecipientsPolicy` never inspects it and the unbounded `transfers` array has no cap. Adding it via `additionalTools` throws `Unhandled tool` at `:66`.

Params use tokenTransfers with a negative sender entry, same shape as the allowance sibling, so `_countTokenAllowance` works as-is: rename it `_countTokenTransfers` and map both tools. Plus a test, 6 recipients with `maxRecipients: 5` should block.

kierzniak (changes requested):

packages/core/src/shared/hedera-utils/hedera-parameter-normaliser.ts
skurzyp-blockydevs: You are right, thanks

docs/HEDERATOOLS.md
skurzyp-blockydevs: Oh, good catch. Forgot to update those. Thanks!

kierzniak (approved):

PR #1057, opened 2026-07-29
feat: add tool type field and `BaseQueryTool` class

### LINKED ISSUE #1002: Feature: Add `toolType` enum field to `BaseTool`
**Related issue:** https://github.com/hashgraph/hedera-agent-kit-js/issues/861

## Problem

Issue #861 requests a `readOnly: true` toolkit option so consumers can declaratively prevent write tools from being handed to an LLM agent. The current workaround is a post-filter on `toolkit.getTools()` using a `get_*` name prefix regex — fragile, silent on new tools, and provides no type-system guarantee.

Name-based filtering is the root problem: it leaks the "is this tool safe?" contract into every consumer instead of encoding it in the tool itself.

## Proposal

Add a `toolType` field to the `Tool` interface and `BaseTool`:

```ts
export type ToolType = 'query' | 'transaction' | 'other';

export interface Tool {
  // ...existing fields...
  toolType: ToolType;
}
```

Default in `BaseTool` is `'other'` so existing custom tools remain backwards-compatible without any changes.

### Intermediate abstract classes (preferred over per-tool manual assignment)

Rather than requiring every built-in tool to manually set `toolType`, introduce two thin subclasses:

```ts
export abstract class BaseTransactionTool extends BaseTool {
  readonly toolType: ToolType = 'transaction';
}

export abstract class BaseQueryTool extends BaseTool {
  readonly toolType: ToolType = 'query';
}
```

All existing transaction tools move from `extends BaseTool` → `extends BaseTransactionTool`.  
All existing query tools move from `extends BaseTool` → `extends BaseQueryTool`.  
No per-tool boilerplate needed.

### Benefits

- **Filtering without name heuristics** — consumers filter by intent, not string prefix:
  ```ts
  toolkit.getTools().filter(t => t.toolType === 'query');
  ```
- **Future-proof** — new write tools added to any plugin are automatically typed; no consumer regex update needed.
- **Policy integration** — policies can target a type instead of a name list:
  ```ts
  class ReadOnlyPolicy implements Policy {
    check(tool: Tool) {
      if (tool.toolType !== 'query') throw new PolicyViolationError(...);
    }
  }
  ```
- **Docs example** — a short snippet showing how to build a read-only agent satisfies the original request from #861 without needing a dedicated `readOnly` flag or `coreReadOnlyPlugin` bundle.

## Scope

1. Add `ToolType` enum and `toolType` field to `Tool` interface (`packages/core/src/shared/tools.ts`)
2. Add `BaseTransactionTool` and `BaseQueryTool` abstract subclasses in the same file
3. Migrate all built-in tools to the appropriate subclass
4. Export new types from `packages/core/src/index.ts`
5. Add a docs example showing query-only filtering and policy usage

skurzyp-blockydevs:
**Description**:

**Related issue(s)**:

Fixes #1002 

**Notes for reviewer**:

**Checklist**


docs/HOOKS_AND_POLICIES.md
```diff
@@ -732,8 +732,58 @@ protected shouldBlockPreToolExecution(
...
+export class ReadOnlyPolicy extends AbstractPolicy {
+  name = 'ReadOnlyPolicy';
+
+  protected shouldBlockPreToolExecution(
+    params: PreToolExecutionParams,
+    _method: string,
+  ): boolean {
+    return params.tool.toolType !== TOOL_TYPE.QUERY;
+  }
+}
```
kierzniak, line 805: The policy reads params.tool.toolType. Where params is `PreToolExecutionParams` but it has no tool property:

```
export interface PreToolExecutionParams<TParams = any> {
  context: Context;
  rawParams: TParams;
  client: Client;
}
```

docs/HEDERATOOLS.md
```diff
@@ -4,6 +4,32 @@ This document provides detailed parameter specifications and example prompts for
...
+| `'transaction'` | `BaseTransactionTool`   | State-mutating — builds and (depending on `AgentMode`) submits a Hedera transaction. |
+| `'other'`       | `BaseTool` (direct)     | Custom tool that does not fit either category; default for tools that extend `BaseTool` directly. |
+
+Use `TOOL_TYPE` constants for safe comparisons. The most common use-case is building a **read-only agent** that can never submit transactions:
+
+```ts
+import { TOOL_TYPE } from '@hashgraph/hedera-agent-kit';
+
+// Only hand query tools to the LLM — no transaction tools included.
+const readOnlyTools = toolkit.getTools().filter(t => t.toolType === TOOL_TYPE.QUERY);
```
kierzniak, line 25: `toolkit.getTools()` returns adapter-built copies that don't carry toolType, so the read-only filter matches nothing and you end up with zero tools.

kierzniak (changes requested):

packages/core/src/shared/hook.ts
```diff
@@ -55,28 +64,40 @@ export abstract class AbstractHook {
...
+   * This is useful for type-based hooks/policies that operate on `params.toolType` rather than
+   * on a specific method name.
    */
   public abstract relevantTools: string[];
 
+  /**
+   * Returns `true` when this hook should run for the given method.
+   * The special sentinel `'*'` in {@link relevantTools} matches any method.
+   */
+  protected appliesToMethod(method: string): boolean {
```
MWBlocky, line 95: The JSDoc promises that relevantTools = ['*'] applies the hook "to every tool", but appliesToMethod is protected and only called from the base no-op implementations (:83, :90, :94, :101); any hook that overrides those methods never sees the wildcard - which is the case for both hooks shipped in this repo (hcs-audit-trail-hook, hol-audit-trail-hook), so ['*'] silently logs nothing.

fix:
Either enforce the wildcard centrally or scope the JSDoc to say '*' only works for hooks that call appliesToMethod() in their overrides.

packages/core/src/shared/tools.ts
MWBlocky, line 198: Filtering by relevantTools is spread across hook implementations — the loop invokes every hook unconditionally, so the '*' contract declared in hook.ts is opt-in per hook author.

this is the root cause of packages/core/src/shared/hook.ts:68 

fix:
 Skip the hook inside executeHooks unless hook.relevantTools contains '*' or this.method, so no subclass can silently ignore the contract.

docs/HOOKS_AND_POLICIES.md
```diff
@@ -735,8 +739,75 @@ protected shouldBlockPreToolExecution(
...
+
+// Attach to context:
+const context: Context = {
+  hooks: [new ReadOnlyPolicy()],
+  // ...
+};
 ```
 
+Both approaches are complementary: Option A reduces the tool list seen by the LLM; Option B
+adds a runtime guardrail that will block even if a transaction tool slips through.
```
MWBlocky: "adds a runtime guardrail that will block even if a transaction tool slips through" is untrue for legacy object-literal tools, which never go through hooks at all, so ReadOnlyPolicy cannot stop them - a security-adjacent overclaim.

fix:
 Qualify the sentence: the policy only guards tools extending BaseTool;  object-literal tools bypass hooks entirely.

docs/HEDERATOOLS.md
```diff
@@ -4,6 +4,45 @@ This document provides detailed parameter specifications and example prompts for
...
+```ts
+import { TOOL_TYPE } from '@hashgraph/hedera-agent-kit';
+
+const allTools = toolkit.getTools();
+const readOnlyTools = Object.fromEntries(
+  Object.entries(allTools).filter(([, t]) => t.toolType === TOOL_TYPE.QUERY),
+);
+```
+
+`toolType` is also included in the `ToolSummary` returned by `HederaAgentAPI.listTools()` — this works regardless of which adapter you are using:
```
MWBlocky: "this works regardless of which adapter you are using" does not hold for the MCP adapter: HederaMCPToolkit keeps its HederaAgentAPI in a private field and registers tools without any type information, so an MCP consumer has no access to toolType.

docs/MIGRATION-v4.md
```diff
@@ -806,19 +809,24 @@ export default tool;
...
 | Hook/Policy support | ✗ None                                       | ✓ Automatic at stages 1, 3, 5, 7                              |
 | Error handling      | Manual `try/catch` inside `execute`          | Override `handleError()` (BaseTool provides a default)        |
 | Breaking change?    | —                                            | **No** — `BaseTool implements Tool`                           |
 
 ### Migrating a query-only tool
 
-For tools that only read data (no transaction signing), override `shouldSecondaryAction` to skip stage 6:
+For tools that only read data (no transaction signing), you can extend `BaseTool` directly
+(and override `shouldSecondaryAction` to return `false`), or use the `BaseQueryTool`
+convenience subclass which sets `toolType = 'query'` for you. Both are valid — `BaseQueryTool`
```
MWBlocky: "Both are valid - BaseQueryTool is just a helper that saves a line" is misleading: extending BaseTool leaves toolType='other', so such a read-only tool is dropped by === TOOL_TYPE.QUERY filters and blocked by the ReadOnlyPolicy introduced in this same PR.

packages/core/tests/unit/shared/api.unit.test.ts
```diff
@@ -4,6 +4,10 @@ import { z } from 'zod';
 
 import HederaAgentAPI from '../../../src/shared/api';
 import type { Tool } from '../../../src/shared/tools';
+import { BaseTool, TOOL_TYPE } from '@/shared';
+import { BaseQueryTool } from '@/shared/base-query-tool';
```
MWBlocky: Add export * from './base-query-tool'; to src/shared/index.ts so both base  classes resolve from @/shared.

docs/HEDERATOOLS.md
MWBlocky, line 912: The branch added a **Tool type** row to 43 of 44 tool sections — TRANSFER_FUNGIBLE_TOKEN_TOOL was skipped.

MWBlocky (approved):

kierzniak (approved):

PR #876, opened 2026-05-30
docs(plugins): add hak-github-pay-plugin v2.3.0 — pay GitHub contributors in HBAR on merge

jmgomezl:
Adds **hak-github-pay-plugin** to the third-party plugins list in `docs/PLUGINS.md` and `README.md`.

### What it does
A HAK v4 plugin that turns the GitHub merge button into a payment rail: when a pull request is merged, an AI agent pays the contributor in HBAR. There is no database — payment policy, the GitHub→Hedera identity registry, every receipt, and software-release provenance all live immutably on four Hedera Consensus Service topics, verifiable on Hashscan.

### Tools (7)
- `github_pay_register_contributor` — map a GitHub handle → Hedera account on a public IDENTITIES topic (a reusable public good)
- `github_pay_set_payment_policy` — `label → HBAR` payout rules
- `github_pay_set_payment_cap` — monthly + per-contributor spending ceilings
- `github_pay_pay_on_merge` — HMAC-validated, idempotent, cap-enforced payout + sealed receipt
- `github_pay_query_contributor_payments` — payment history as CSV
- `github_pay_seal_release_provenance` — SHA-256 release-asset provenance (NIST SSDF)
- `github_pay_query_team_summary` — per-contributor aggregate CSV

### Links
- NPM: https://www.npmjs.com/package/hak-github-pay-plugin
- Repo: https://github.com/jmgomezl/hak-plugin-github-pay
- Live demo: https://github-pay.aivylabs.xyz

Built on Hedera Agent Kit v4 (`BaseTool` + plain-object plugin pattern, config resolved from the HAK `Context`). Commits are DCO signed-off.

PR #828, opened 2026-05-19
docs(plugins): add hak-axelar-plugin v1.0.0 — Axelar cross-chain bridging on Hedera

jmgomezl:
## Summary

Adds **[hak-axelar-plugin](https://www.npmjs.com/package/hak-axelar-plugin)** to the third-party plugins list in both `README.md` and `docs/PLUGINS.md`.

### Plugin overview

`hak-axelar-plugin` enables AI agents to bridge ITS tokens and send GMP cross-chain messages between Hedera and 60+ chains via the [Axelar Amplifier](https://axelar.network/) architecture.

**Tools exposed:**
- `axelar_get_supported_chains` — list chains available via the Axelar GMP API
- `axelar_get_message_fee` — estimate relay gas cost for a destination chain
- `axelar_send_message` — send an arbitrary GMP payload to an AxelarExecutable contract on another chain (two-tx flow: GasService + Gateway)
- `axelar_send_token` — bridge a registered ITS token in a single transaction via `InterchainTokenService.interchainTransfer()`
- `axelar_get_message_status` — poll delivery status from the Axelar GMP API

**Notes:**
- Uses Hedera's Amplifier architecture (not classic gateway) — deposit-address bridging and axlUSDC are not available; only ITS-registered tokens can be bridged
- No `@axelar-network/axelarjs-sdk` dependency — direct REST calls to avoid ethers v5 / ESM conflicts
- v4-compatible: plain `Plugin` object, zod v3, `BaseTool` with `normalizeParams` / `coreAction` / `secondaryAction`

**Links:**
- NPM: https://www.npmjs.com/package/hak-axelar-plugin
- GitHub: https://github.com/jmgomezl/hak-axelar-plugin

## Checklist

- [x] Entry added to `README.md` Third Party Plugins section
- [x] Entry added to `docs/PLUGINS.md` with NPM, GitHub, version, and status fields
- [x] Commit signed off (`-s`) per DCO requirement
- [x] Single clean commit

PR #823, opened 2026-05-16
docs: add hak-layerzero-plugin to third-party plugins list

jmgomezl:
## Summary

- Adds **hak-layerzero-plugin@1.0.0** entry to `docs/PLUGINS.md` under Available Third Party Plugins
- Adds the same plugin to the **Third Party Plugins** section of `README.md`

The plugin enables AI agents to send cross-chain messages and query fees between Hedera and 150+ EVM chains via [LayerZero V2](https://layerzero.network/) — filling the cross-chain gap in the Hedera DeFi agent stack.

**Published package:** https://www.npmjs.com/package/hak-layerzero-plugin  
**GitHub repo:** https://github.com/jmgomezl/hak-layerzero-plugin

### Tools exposed

| Tool | Type | Description |
|---|---|---|
| `layerzero_get_supported_chains` | Query | Lists chains reachable from Hedera with their LayerZero EIDs |
| `layerzero_get_message_fee` | Query | Estimates HBAR fee via on-chain `EndpointV2.quote()` |
| `layerzero_send_message` | Transaction | Sends cross-chain message from OApp on Hedera |
| `layerzero_get_message_status` | Query | Tracks delivery status via LayerZero Scan API |

### Technical details

- TypeScript ESM + CJS dual output, HAK v4 `BaseTool` architecture
- Confirmed Hedera mainnet EndpointV2: `0x3A73033C0b1407574C76BdBAc67f126f6b4a9AA9` (EID 30316)
- 25 unit tests, all passing
- Full docs: TOOLS.md, CONFIGURATION.md, EXAMPLES.md, AGENTS.md

> **Note:** Assignee Check CI failure is expected for external contributors — maintainers assign themselves.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

PR #822, opened 2026-05-16
docs(plugins): add hak-stader-plugin v1.0.0 — Stader liquid staking on Hedera

jmgomezl:
## Summary

Adds [`hak-stader-plugin@1.0.0`](https://www.npmjs.com/package/hak-stader-plugin) to the third-party plugins list in `docs/PLUGINS.md` and `README.md`.

This plugin integrates [Stader](https://www.staderlabs.com/hedera/) liquid staking on Hedera, allowing agents to stake HBAR and receive **HBARX** — a yield-bearing liquid staking token that remains transferable and usable in DeFi while earning staking rewards.

## Tools exposed

| Tool | Type | Description |
|------|------|-------------|
| `stader_stake_hbar` | Transaction | Stake HBAR, receive HBARX |
| `stader_approve_hbarx` | Transaction | Grant HTS allowance before unstaking |
| `stader_unstake_hbar` | Transaction | Initiate unstake (1-day unbonding period) |
| `stader_claim_withdrawal` | Transaction | Claim HBAR after unbonding |
| `stader_get_pending_withdrawals` | Query | List pending withdrawal requests with claimable status |
| `stader_get_hbarx_balance` | Query | Get HBARX token balance for an account |
| `stader_get_exchange_rate` | Query | Get HBARX/HBAR rate computed via Hedera Mirror Node |
| `stader_get_staking_info` | Query | Protocol status, unbonding period, contract addresses |

## Implementation notes

- Built on `BaseTool` (v4 pattern) — full hooks and policies support
- Exchange rate computed off-chain via Mirror Node (the on-chain `getExchangeRate()` is deprecated and always returns 1.0)
- Unstaking requires a prior HTS token allowance (`AccountAllowanceApproveTransaction`) — the plugin exposes a dedicated `stader_approve_hbarx` tool for this
- 50 unit tests, zero real network calls in tests (all dependencies injected via context)
- Pairs naturally with `hak-saucerswap-plugin` for a full agent-driven yield loop: stake HBAR → receive HBARX → LP on SaucerSwap → farm rewards

## Links

- NPM: https://www.npmjs.com/package/hak-stader-plugin
- GitHub: https://github.com/jmgomezl/hak-stader-plugin
- Version: `hak-stader-plugin@1.0.0`

## Checklist

- [x] Entry added to `docs/PLUGINS.md` under Available Third Party Plugins
- [x] Entry added to `README.md` under Third Party Plugins
- [x] Commit is DCO-signed (`git commit -s`)
[...]

PR #1068, opened 2026-08-03
docs: endorse hak-pyth-plugin@0.3.0 and flag the Pyth Core API-key deadline

jmgomezl:
## What

Updates the endorsed version of `hak-pyth-plugin` to **0.3.0** in both places it is listed, and adds a time-sensitive upgrade note.

## Why

**1. The two lists were stale and disagreed with each other.**

| File | Was | Now |
|---|---|---|
| `docs/PLUGINS.md` | `0.2.0` | `0.3.0` |
| `packages/core/README.md` | `0.1.1` | `0.3.0` |

**2. There is a hard deadline attached to this one.**

Pyth Core requires an API key from **2026-08-18** under [OP-PIP-100](https://forum.pyth.network/t/passed-op-pip-100-pyth-core-to-pyth-pro-migration/2420). From that date the Hermes endpoint rejects anonymous price-update requests, so **every published version below 0.3.0 stops returning prices.** This is not a deprecation warning — the tools start failing.

I verified the enforcement is already live on the upgraded endpoint:

| Endpoint | `/v2/price_feeds` | `/v2/updates/price/latest` |
|---|---|---|
| `hermes.pyth.network` (current) | 200 | 200 |
| `pyth.dourolabs.app/hermes` (upgraded) | 200 | **401** |

0.3.0 adds API-key support and selects the endpoint accordingly: without a key it stays on the current Hermes host (works until 2026-08-18), and setting `PYTH_API_KEY` switches it to the upgraded authenticated endpoint. Upgrading is non-breaking — existing users keep working, and setting the env var is the whole migration.

Plugin-side change: [jmgomezl/hak-pyth-plugin#5](https://github.com/jmgomezl/hak-pyth-plugin/pull/5).

## Scope

Docs only — two version strings and two upgrade notes. No code changes.

Note: both files already fail `prettier --check` on `main` today, before this change. I left the existing formatting untouched so the diff stays reviewable rather than reformatting whole files in a docs PR.

### DIFF
--- a/docs/PLUGINS.md
+++ b/docs/PLUGINS.md
@@ -60,9 +60,16 @@ See this list of available third party plugins for the Hedera Agent Kit in the [
 
   Github repository: https://github.com/jmgomezl/hak-pyth-plugin
 
-  Version: hak-pyth-plugin@0.2.0
+  Version: hak-pyth-plugin@0.3.0
 
   Status: Validated by HAK team, v4-compatible release
+
+  > **Upgrade required before 2026-08-18.** Pyth Core now requires an API key
+  > ([OP-PIP-100](https://forum.pyth.network/t/passed-op-pip-100-pyth-core-to-pyth-pro-migration/2420)).
+  > From that date the Hermes endpoint rejects anonymous price requests, so versions
+  > below 0.3.0 stop returning prices. Upgrade to 0.3.0 and set `PYTH_API_KEY`
+  > ([get a key](https://docs.pyth.network/price-feeds/pro/acquire-api-key)); the plugin
+  > then targets the upgraded endpoint automatically.
 
 
 - [CoinCap Plugin](https://www.npmjs.com/package/coincap-hedera-plugin) provides access to the [**CoinCap API service**](https://www.coincap.io) to access cryptocurrency market data. It exposes the action (`get HBAR price in USD`) to get the current price of HBAR in USD currency, by using it you can ask your agent to get your current HBAR balance expressed in USD.
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -81,7 +81,11 @@ The Hedera Agent Kit is extensible with third party plugins by other projects. S
 - [Pyth Plugin](https://www.npmjs.com/package/hak-pyth-plugin) provides access to the [**Pyth Network**](https://www.pyth.network/) price feeds via the Hermes API, exposing tools to list feeds and fetch latest prices:
 
   Github repository: [https://github.com/jmgomezl/hak-pyth-plugin](https://github.com/jmgomezl/hak-pyth-plugin).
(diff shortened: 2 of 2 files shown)


jmgomezl:
Flagging a CI issue I can't resolve from my side: **`Assignee Check` is failing, and it needs a maintainer.**

The `title-check` workflow fails when a PR has no assignee:

```yaml
- name: Check Assignee
  if: ${{ github.event.pull_request.assignees == null || github.event.pull_request.assignees[0] == null }}
  run: |
    echo "Assignee is not set. Failing the workflow."
    exit 1
```

I tried to self-assign when opening the PR, but outside contributors don't have the permission:

```
GraphQL: jmgomezl does not have the correct permissions to execute `ReplaceActorsForAssignable`
```

Assigning requires write access (or being a repo collaborator), so this check can't pass on a fork-based PR until someone on the team sets an assignee. It looks like it would affect any external contribution, not just this one — worth a look if that isn't intended.

Every other check is green: DCO, Title Check, conventional-pr-title, all five Wiz scanners, and Snyk license + security.

Could a maintainer assign this one so CI can go green? Happy to rebase if anything else is needed.

jmgomezl:
@MWBlocky @skurzyp-blockydevs — could one of you set an assignee on a few PRs? Sorry to ping directly; this is the one step I can't do from a fork.

**`Assignee Check` is the only failing check on five of my open PRs.** Every other check passes on all of them. Outside contributors can't self-assign — GitHub rejects it:

```
GraphQL: jmgomezl does not have the correct permissions to execute `ReplaceActorsForAssignable`
```

You assigned #822 and #828 previously, @MWBlocky, which is exactly why those two now go green — so I think this is just the same one-click action, five times.

| PR | Files | Status |
|---|---|---|
| #1068 | 2 | Endorses `hak-pyth-plugin@0.3.0`. **Time-sensitive** — Pyth Core requires an API key from 2026-08-18, and versions below 0.3.0 stop returning prices on that date. |
| #921 | 1 | Adds `hak-uniswap-plugin` to the plugin list. |
| #922 | 1 | Adds `hak-ledger-plugin` to the plugin list. |
| #876 | 1 | Adds `hak-github-pay-plugin`. Rebased today. |
| #893 | 9 | `swap_tokens_tool` (core-dex-plugin). Rebased today; Wiz SAST now clean after a validation fix. |

All five are `MERGEABLE` against current `main` — I rebased the stale ones today and dropped their `README.md` entries, since the per-plugin list has been consolidated into `docs/PLUGINS.md`. #822, #823 and #828 are also green and ready.

No rush on the review itself — I know the queue is long. It's only the assignment that's blocking CI.

One thing worth flagging separately: this check appears to fail for **any** fork-based PR, since no outside contributor can self-assign. If that isn't intended, it may be worth exempting external PRs or having the workflow auto-assign the author. Happy to open a separate issue if that's useful.

